PR #15, opened 2021-05-16
remove vite/dynamic-import-polyfill

shaal:
When running storybook, I saw this message in terminal
```
'vite/dynamic-import-polyfill' is no longer needed, refer to https://github.com/vitejs/vite/blob/main/packages/vite/CHANGELOG.md#230-2021-05-10
```


eirslett:
Thanks!

PR #17, opened 2021-05-25
Simple manual CI

sevenfields-matt:
Here is a simple version of the CI.

What it does:

1. Should push to npm if you define a secret called:  NPM_TOKEN

What it doesn't do:

1. Version.  I wasn't sure how you wanted to do it.  So for now you could just run  
`npm version` 
or 
`yarn version` 

2.  Set the npm package public.  (But you can add this by adding --access public to the npm publish command)

As this PR sits you could use it like this:

1. Decide to release
2. npm version
3. push
4. run CI
5. Set package to public when you wanted to.

### DIFF
--- a/.github/workflows/buildAndPublish.yml
+++ b/.github/workflows/buildAndPublish.yml
@@ -1,0 +1,35 @@
+name: Node CI
+
+on: [workflow_dispatch]
+
+defaults:
+  run:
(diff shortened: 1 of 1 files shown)


eirslett:
Thanks a lot! 🥳

PR #22, opened 2021-06-09
Move import of framework code to top

### LINKED ISSUE #19: Accessing non-existent addons channel
Hi, I'm just switching over to vite, and trying out this builder, and while it seems to build correctly in the terminal, when storybook opens I see an error in the console, and the page does not finish loading:

```
Uncaught Error: Accessing non-existent addons channel, see https://storybook.js.org/basics/faq/#why-is-there-no-addons-channel
    at AddonStore2.getChannel (:6006/@fs/Users//.../node_modules/storybook-builder-vite/node_modules/.vite/chunk-LAF5HDGZ.js?v=485b51f5:365)
    at :6006/@fs/Users/.../.storybook/preview.js:18
```

I am using storybook-dark-mode, and listening to events from that addon using their instructions like: https://github.com/hipstersmoothie/storybook-dark-mode#events.  This worked fine with the default storybook webpack builder.  

Edit: when I remove that addon, I get a similar error from addon-a11y.

I am using storybook 6.3.0-beta.2

IanVS:
Fixes #19 

The import of `@storybook/<framework>` needs to occur first thing in the iframe, to set up the store and addons channel and whatnot.  It was [these lines](https://github.com/storybookjs/storybook/blob/0981b635eed95b48998b9a5fd6d67216802324fb/lib/builder-webpack4/src/preview/iframe-webpack.config.ts#L90-L91) in the webpack builder which tipped me off (and which I stole here).

So, this moves up the import to the top, and also allows the use of a custom frameworkPath, the same way the webpack builder does.  

I've confirmed this fixes my problem about the addons channel as well.

### DIFF
--- a/packages/storybook-builder-vite/codegen-iframe-script.js
+++ b/packages/storybook-builder-vite/codegen-iframe-script.js
@@ -13,11 +13,16 @@ function replaceCJStoESMPath(entryPath) {
 module.exports.generateIframeScriptCode = async function generateIframeScriptCode(
     options
 ) {
-    const { presets, configDir, framework } = options;
+    const { presets, configDir, framework, frameworkPath } = options;
 
     const previewEntries = (
         await presets.apply('previewEntries', [], options)
     ).map(replaceCJStoESMPath);
+
+    // Ensure that the client API is initialized by the framework before any other iframe code
+    // is loaded. That way our client-apis can assume the existence of the API+store
+    const frameworkImportPath = frameworkPath || `@storybook/${framework}`;
+
     const configEntries = [loadPreviewOrConfigFile({ configDir })]
         .concat(await presets.apply('config', [], options))
         .filter(Boolean);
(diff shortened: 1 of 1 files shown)


packages/storybook-builder-vite/codegen-iframe-script.js
```diff
@@ -13,11 +13,16 @@ function replaceCJStoESMPath(entryPath) {
 module.exports.generateIframeScriptCode = async function generateIframeScriptCode(
     options
 ) {
-    const { presets, configDir, framework } = options;
+    const { presets, configDir, framework, frameworkPath } = options;
```
eirslett, line 16: Does frameworkPath come as a config option from Storybook core? What's the use case for being able to override it?

IanVS: The use case is discussed in https://github.com/storybookjs/storybook/issues/12075, and support in webpack was added in https://github.com/storybookjs/storybook/pull/12087.  I _think_ these are the same options that the webpack builder also gets from storybook, right?

eirslett: Ah, I see. Thanks!

PR #28, opened 2021-06-11
Tweak Vite setup so we use the correct working directory

eirslett:
We use a heuristic: assume that the working directory/project root for Storybook is the parent directory of options.configDir. I don't think there's a config variable to specify the exact root directory. This lets us tighten security by turning on fsserve.strict.

https://vitejs.dev/config/#server-fsserve-strict

If the heuristic is not good enough to detect people's working directories, they can override it in viteFinal().

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,12 @@ The `configType` variable will be either `"DEVELOPMENT"` or `"PRODUCTION"`.
 
 The function should return the updated Vite configuration.
 
+## Note about working directory
+
+The builder will by default enable Vite's [server.fsServe.strict](https://vitejs.dev/config/#server-fsserve-strict)
+option, for increased security. The default `server.fsServe.root` is set to the parent directory of the
+storybook configuration directory. This can be overridden by configuring fsServe in viteFinal.
+
 ### Getting started with React, Vite and Storybook (on a new project)
 
 ```
--- a/README.md
+++ b/README.md
@@ -59,8 +59,8 @@ The function should return the updated Vite configuration.
 ## Note about working directory
 
 The builder will by default enable Vite's [server.fsServe.strict](https://vitejs.dev/config/#server-fsserve-strict)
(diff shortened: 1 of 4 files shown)


IanVS (changes requested):
I've confirmed that the default cache is now being placed in my node_modules/.vite, as I would expect.  That alone makes this a big win, I think.  

What is the best way to test that the added security of using fsserve.strict is working as intended?

And, I think that before we merge this, there should be some updates to the readme to mention this behavior (the heuristic) and suggest ways to override it if necessary.  I can see this being the kind of "magic" that could cause problems if people don't know about it.

eirslett:
> What is the best way to test that the added security of using fsserve.strict is working as intended?

I think we have to test it manually. I tried running Storybook locally with fsserve.strict disabled - I could some files in my filesystem that were no longer possible to access after enabling the setting.

> And, I think that before we merge this, there should be some updates to the readme to mention this behavior (the heuristic) and suggest ways to override it if necessary.

Good point, I added some docs now!

README.md
```diff
@@ -56,6 +56,12 @@ The `configType` variable will be either `"DEVELOPMENT"` or `"PRODUCTION"`.
 
 The function should return the updated Vite configuration.
 
+## Note about working directory
+
+The builder will by default enable Vite's [server.fsServe.strict](https://vitejs.dev/config/#server-fsserve-strict)
+option, for increased security. The default `server.fsServe.root` is set to the parent directory of the
```
IanVS, line 62: It looks like `root` is being set at the root of the config, not nested under `server.fsServe`, isn't it?  Shouldn't they be changing their top-level root if we guess it wrong?

eirslett: You're right! There are two `root` config options (one top-level and one under server.fsServe) and they behave a bit differently. I updated the PR.

IanVS (approved):

Issue #31, opened 2021-06-17
Can't build the example with React


rqbazan:
### Reproduction: 

1. Clone this repo: `git clone https://github.com/eirslett/storybook-builder-vite.git`
2. Go to the React example folder: `cd storybook-builder-vite/packages/example-react`
3. Install packages: `npm i` 
4. Try to build the storybook app: `npm run build-storybook` 💥🐛

### Logs: 
```node
🚀 npm run build-storybook

> example-react@0.0.7 build-storybook D:\Software\Github\storybook-builder-vite\packages\example-react
> build-storybook

info @storybook/react v6.3.0-rc.10
info
info => Cleaning outputDir: D:\Software\Github\storybook-builder-vite\packages\example-react\storybook-static
info => Loading presets
ERR! Addon value should end in /register OR it should be a valid preset https://storybook.js.org/docs/react/addons/writing-presets/
ERR! @storybook/addon-links
vite v2.3.7 building for production...
transforming (1154) node_modules\@popperjs\core\lib\dom-utils\isTableElement.jsConflicting namespaces: "node_modules\@storybook\client-api\dist\esm\index.js" re-exports "combineParameters" from both "node_modules\@storybook\client-api\dist\esm\index.js" and "node_modules\@storybook\client-api\dist\esm\parameters.js" (will be ignored)
Conflicting namespaces: "node_modules\@storybook\components\dist\esm\index.js" re-exports "components" from both "node_modules\@storybook\components\dist\esm\index.js" and "node_modules\@storybook\components\dist\esm\typography\DocumentFormatting.js" (will be ignored)
Conflicting namespaces: "node_modules\@storybook\addon-docs\dist\esm\blocks\index.js" re-exports "SourceState" from both "node_modules\@storybook\addon-docs\dist\esm\blocks\Canvas.js" and "node_modules\@storybook\addon-docs\dist\esm\blocks\Source.js" (will be ignored)
Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification
Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification
Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification
✓ 1180 modules transformed.
rendering chunks (7)...[vite:build-html] The "fileName" or "name" properties of emitted files must be strings that are neither absolute nor relative paths, received "../../../../../../iframe.html".
[...]

PR #34, opened 2021-06-18
Add build examples workflow

rqbazan:
### Context

Please read https://github.com/eirslett/storybook-builder-vite/issues/31

### Changes

My very first idea was to add a new step in the current `Build and Publish to NPM` workflow in order to build all the [examples packages](https://github.com/eirslett/storybook-builder-vite/tree/3abcdd1d4cd1c3f82f23e3e636c4716fbe55765e/packages). However, I've decided to create a new workflow because the build examples step must run three times, one for each operative system, so this behavior isn't properly along with a publish step.


IanVS:
It seems there's a problem running the actions:

```
GitHub Actions has encountered an internal error when running your job.
```

rqbazan:
Hey @IanVS, please re-run

eirslett:
Looks like it fails on Windows! Maybe something with `/` vs `\` in file paths?

rqbazan:
Hey @eirslett,

That's sounds right because the error message shows a relative path but it seems unrecognizable by the build process. 🤔 

```node
 ERR! Error: The "fileName" or "name" properties of emitted files must be strings that are neither absolute nor relative paths, received "../../../../../../iframe.html".
```

IanVS:
I think this properly reproduces the issue in https://github.com/eirslett/storybook-builder-vite/issues/31, and adds value to our build process.  We (or someone!) will need to spend some work to fix the windows issue, but at least now we'll have a reminder that it's broken and a good way to check if it's fixed.  Thanks @rqbazan!

PR #42, opened 2021-06-23
Do not depend on framework plugins (mostly)

### LINKED ISSUE #41: Remove some dependencies
Adding the storybook framework addons and plugins was a mistake, I think.  I've tried out the latest version of this builder, and now storybook thinks I'm a svelte project, when really I'm using react.  It seems this is because storybook-builder-vite depends on `@storybook/addon-svelte-csf`, which in turn brings in `@storybook/svelte`, and storybook uses the presence of a framework library to determine what to run, and it blows up in my react project.

I'm kind of wondering if maybe we should leave it up to the consumer to add the plugins that they need for their framework.  We can document the recommendations for the various frameworks, but then leave it up to users to add them.  It's likely they'll already be using them in their production app already, after all.

The other possibility is to try to read their vite.config file and automatically extend that.  But that's maybe a bigger step than we should try taking right now. 

Or, we can remove the dependencies from our package.json, but still try to require them and add them in the config, maybe with a try/catch to throw a friendly message if they're not found.  I'd suggest peerDependencies, but that's turned into a complete dumpster fire.

IanVS:
Closes #41 

This removes the direct dependency on a few plugins, most notably @storybook/addon-svelte-csf, which causes @storybook/svelte to also be installed, which can cause storybook to run in svelte mode when it shouldn't.

This takes the approach of moving vue and svelte addons/plugins to devDependencies and throwing a friendly error message if they are attempted to be required but are not found.

I did not give react the same treatment, because I am planning to publish a forked version of plugin-react-refresh which should allow us to restart the builder when stories change, so that changes at least appear, even if they're not HMR'ed.

### DIFF
--- a/packages/storybook-builder-vite/package.json
+++ b/packages/storybook-builder-vite/package.json
@@ -14,13 +14,15 @@
     "homepage": "https://github.com/eirslett/storybook-builder-vite/#readme",
     "dependencies": {
         "@mdx-js/mdx": "^1.6.22",
-        "@storybook/addon-svelte-csf": "^1.1.0",
         "@storybook/csf-tools": "^6.3.0-rc.4",
-        "@sveltejs/vite-plugin-svelte": "^1.0.0-next.7",
         "@vitejs/plugin-react-refresh": "^1.3.2",
-        "@vitejs/plugin-vue": "^1.2.1",
         "glob-promise": "^4.1.0",
         "vite-plugin-mdx": "^3.5.1"
+    },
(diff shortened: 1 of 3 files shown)


eirslett (approved):

eirslett:
@IanVS Did this PR break the example packages? Do we need to update their package.json to include the correct dependencies?

IanVS:
I'll fix them up. I wonder why CI passed on this PR originally...

PR #51, opened 2021-06-28
Ensure examples build in CI

IanVS:
The main problem this addresses is that we need to ensure that `yarn install` is run no matter what, even if the cache is restored.  This is because yarn relies on having an `install-state.gz` file that is gitignored and created during installations.  We were skipping the installation entirely before if there was a cache, which is the reason that https://github.com/eirslett/storybook-builder-vite/pull/45 is failing in CI.  

This also adds a missing dependency, `@sveltejs/vite-plugin-svelte` to the svelte example.  It would be yarn installed along with the rest of the workspace, but for folks using npm or using the example as a guide, it's important to have it there.

### DIFF
--- a/.github/workflows/buildExamples.yml
+++ b/.github/workflows/buildExamples.yml
@@ -33,7 +33,6 @@ jobs:
 
       - name: Install dependencies
         run: yarn install
-        if: steps.yarn-cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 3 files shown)


IanVS:
I made sure to run the actions twice, so that the cache is hit.

PR #54, opened 2021-06-28
Sort stories by their export order

schalk-b:
Fix for #38

Needs thought around how we keep track of stories, instead of having to hardcode `/\.stories\.(t|j)sx?$/.test(id)`

### DIFF
--- a/packages/storybook-builder-vite/inject-export-order-plugin.js
+++ b/packages/storybook-builder-vite/inject-export-order-plugin.js
@@ -20,7 +20,7 @@ module.exports.injectExportOrderPlugin = function injectExportOrderPlugin(option
       const [, exports] = parse(code);
       if (exports.includes('__namedExportsOrder')) {
         // user has defined named exports already
-        return null;
+        return;
       }
 
       const orderedExports = exports.filter(e => e !== 'default');
--- a/packages/storybook-builder-vite/inject-export-order-plugin.js
+++ b/packages/storybook-builder-vite/inject-export-order-plugin.js
@@ -1,8 +1,8 @@
 const { parse } = require('es-module-lexer');
 
-module.exports.namedExportOrderPlugin = function namedExportOrderPlugin(options) {
+module.exports.injectExportOrderPlugin = function injectExportOrderPlugin(options) {
   return {
-    name: 'storybook-vite-named-export-order-plugin',
+    name: 'storybook-vite-inject-export-order-plugin',
     // This should only run after the typescript has been transpiled
     enforce: 'post',
     async transform(code, id) {
--- a/packages/storybook-builder-vite/inject-export-order-plugin.js
+++ b/packages/storybook-builder-vite/inject-export-order-plugin.js
@@ -1,32 +1,22 @@
 const { parse } = require('es-module-lexer');
 
-module.exports.injectExportOrderPlugin = function injectExportOrderPlugin(options) {
-  return {
+module.exports.injectExportOrderPlugin = {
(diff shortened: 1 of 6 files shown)


schalk-b:
I also think this should be `inject-export-order` instead of `named-export-order`

eirslett:
> Needs thought around how we keep track of stories, instead of having to hardcode /\.stories\.(t|j)sx?$/.test(id)

We don't even need to care about that. We can inject __namedExportsOrder in every source file, regardless of whether it's a Storybook story or not.

packages/storybook-builder-vite/svelte/csf-plugin.js
```diff
@@ -33,13 +33,19 @@ module.exports = {
...
+              .join("', '");
+
             const output = [
                 codeWithoutDefaultExport,
                 `import parser from '${parser}';`,
                 `const __storiesMetaData = parser(${component}, ${JSON.stringify(
                     all
                 )});`,
                 'export default __storiesMetaData.meta;',
+                `export const __namedExportsOrder = ['${namedExportsOrder}'];`,
```
eirslett, line 48: A simpler option is to just
```
`export const __namedExportsOrder = ${JSON.stringify(namedExportsOrder)};`
```

And then `namedExportsOrder` can simply be an array of strings? There's virtually zero runtime overhead, and saves you from some potential formatting mistakes (theoretically speaking), but most of all makes the code a bit easier to read (in my opinion)?

packages/storybook-builder-vite/inject-export-order-plugin.js
```diff
@@ -0,0 +1,32 @@
...
+      }
+
+      const [, exports] = parse(code);
+      if (exports.includes('__namedExportsOrder')) {
+        // user has defined named exports already
+        return;
+      }
+
+      const orderedExports = exports.filter(e => e !== 'default');
+      const exportsArray = `['${orderedExports.join("', '")}']`;
```
eirslett, line 27: see comment about JSON.stringify()

packages/storybook-builder-vite/svelte/csf-plugin.js
schalk-b: Yup agree, this is currently just a copy and paste from https://github.com/modernweb-dev/web/blob/6c5893cc79c91e82f9aec35d3011c6e33ce878cc/packages/dev-server-storybook/src/shared/stories/injectExportsOrder.ts

schalk-b:
It looks like my implementation generates the warning:
`a plugin (storybook-vite-inject-export-order-plugin) was used to transform files, but didn't generate a sourcemap for the transformation. Consult the plugin documentation for help`

Does anyone have experience with sourcemap generation?

schalk-b:
> > Needs thought around how we keep track of stories, instead of having to hardcode /.stories.(t|j)sx?$/.test(id)
> 
> We don't even need to care about that. We can inject __namedExportsOrder in every source file, regardless of whether it's a Storybook story or not.

Isn't __namedExportsOrder only necessary for stories. I don't think it's wise to do unnecessary parsing and injecting in source files  that aren't stories.

eirslett:
> Isn't __namedExportsOrder only necessary for stories. I don't think it's wise to do unnecessary parsing and injecting in source files that aren't stories.

Yup, for now it's only used for stories. The trade-off with only applying this to stories, is that you need to keep track of which files are stories and which aren't. I don't think it's a big deal if we inject the snippet into all source files. (Not 100 % sure about the source map, though... I suppose we could just reuse the same sourcemap and not make any changes to it.)

PR #72, opened 2021-07-18
Allow passing options to vite-plugin-svelte

dimfeld:
Required to use svelte-preprocess for Typescript, PostCSS, etc. support inside Svelte components. This uses the same `svelteOptions` key used with other builders.


eirslett:
I suppose that should work - can you add a Svelte section in the README, with an example, to document how this would be used?

dimfeld:
Done. Thanks!

PR #98, opened 2021-09-22
Add storybook 6.4 deps to optimizeDeps

IanVS:
This sorts the dependencies, because that was driving me crazy, and adds `lodash/startCase` and `slash` to the list, since they were added in storybook 6.4 it seems.

I think we should also bump our devDependencies to the latest 6.4 beta, but I'm having problems doing so and we can do it in another PR.


eirslett (approved):
Looks good!

PR #101, opened 2021-09-27
Upgrade vite and react plugin

eirslett:
This might be a breaking change, I guess we should bump to version 0.1 (I guess it's time for that now, anyways?)

### DIFF
--- a/packages/storybook-builder-vite/package.json
+++ b/packages/storybook-builder-vite/package.json
@@ -15,7 +15,7 @@
     "dependencies": {
         "@mdx-js/mdx": "^1.6.22",
         "@storybook/csf-tools": "^6.3.3",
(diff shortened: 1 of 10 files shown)


IanVS:
I agree this should be a breaking change, warrants a `0.1.0` release.

eirslett:
...and 0.1.0 is now released!

packages/storybook-builder-vite/package.json
```diff
@@ -24,6 +24,6 @@
         "vue-docgen-api": "^4.40.0"
     },
     "peerDependencies": {
-        "vite": "^2.4.1"
+        "vite": "2.5.10"
```
deek87, line 27: This should be `^2.5.10`, this now locks the peerDependency to 2.5.10
So you can only use this builder with vite 2.5.10 or force the dependency

PR #105, opened 2021-09-29
Inject storybook environment variables

### LINKED ISSUE #104: Environment variables are not injected
I noticed in my project that `import.meta.env.STORYBOOOK` was not defined.  It seems that the webpack builder is using `presets.apply('env');` and this project is not.

IanVS:
Fixes #104.

This injects storybook environment variables through the `define` config of vite, similar to what the webpack builder does using the `DefinePlugin`.  I needed to create a customized version of one of the core storybook functions, since it assumes that `process.env` is going to be used, whereas with vite we use `import.meta.env`.  

I also added example stories in react, vue, and svelte (my first time writing vue and svelte!).  They're ugly, but they show that the STORYBOOK environment variable is being provided correctly (it wasn't previously).

Before:
image

After:
image

### DIFF
--- a/packages/example-react/stories/EnvironmentVariables.jsx
+++ b/packages/example-react/stories/EnvironmentVariables.jsx
@@ -1,0 +1,11 @@
+
+export function EnvironmentVariables() {
+  return (
+    <div>
+      <h1>import.meta.env:</h1>
+      <div>{JSON.stringify(import.meta.env, null, 2)}</div>
+      <h1>import.meta.env.STORYBOOK:</h1>
+      <div>{import.meta.env.STORYBOOK}</div>
+    </div>
+  )
+}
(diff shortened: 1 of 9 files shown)


eirslett:
Do we risk exposing `process.env.SUPER_SECRET_PASSWORD` if that is somehow set up as an environment variable on the server?

IanVS:
No, because storybook only finds variables starting with `_STORYBOOK`.  That's why I'm using preset.apply("env").

eirslett (approved):

FezVrasta:
`import.meta.env.STORYBOOK` (`=== 'true'`) is not set, is this expected?

IanVS:
@FezVrasta I responded to your issue: https://github.com/storybookjs/builder-vite/issues/489

PR #115, opened 2021-10-14
Adding vue2 support

Vnthf:
referencing #66 ,
I fixed infinite loading screen and example and so on from recent main branch. And finished test in both vue2 and vue3


TylerOliver:
I'd like to help make sure #66 or this PR get across the finish line, how can I do that best? Vue 2 support is very important to me as I'm trying to support @vue/composition-api in our component-library, and it works much better with Vite as the builder. I can't migrate to Vue3 until Vuetify goes to Vue 3 as well.

IanVS:
If the issues really are just around adding a vue2 example, I wonder if we should just punt that part of it, and add support for vue2 without an example, with the caveat that it'll be pretty experimental and might break since we won't be testing it.

TylerOliver:
If I put up a PR without an example and with considerations in the documentation, would it be considered @eirslett ?

eirslett:
Sounds OK to me!

Toilal:
Great job @TylerOliver, I have rebased this branch on last commit and it works properly for my project (vite + vue2 + composition api). Would you consider merging this @eirslett ?

IanVS:
@Toilal I don't think we have a PR yet which adds vue2 without an example, as we discussed above.

TylerOliver:
> Great job @TylerOliver, I have rebased this branch on last commit and it works properly for my project (vite + vue2 + composition api). Would you consider merging this @eirslett ?

I haven't done anything @Vnthf did the ground work here. I ended up getting what I needed in the last Vuetify Alpha release to move forward on an alpha, so I'm doing that. I do think there are a lot of enterprises that would love this though.

ushironoko:
The project I belong to also uses vite + vue2 (vite-plugin-vue + @vue/composition-api).
In many cases, companies have been slow to migrate to vue3, and there is a high motivation to maintain the vue2 configuration and leverage the vite ecosystem.

IanVS:
@ushironoko we're not opposed to adding it, but none of the maintainers use vue2, so it's up to someone to submit a PR.  This attempt ran into troubles with the examples within our monorepo, but we've said that we'd be willing to review either this PR or another which adds vue2 without an example.

eirslett:
I've changed my mind a bit on this topic. Take into consideration that this builder is maintained by a few contributors, working as volunteers in their own free time on the project. If none of the maintainers here really use Vue 2, my suggestion is that we don't support Vue 2 at all.

We should try to keep the maintenance burden as low as possible. If we add Vue 2 support, that could break in the future if we make changes to the builder, and without proper testing, it would be hard to detect these bugs.

The best solution, given these human resource constraints, would be that developers with a real need for Vue 2 support take it upon themselves to make their own separate builder, say `storybook-builder-vite-vue2`, and then maintain it in their own GitHub repository and publish it on npm in their own pipeline.

Users of Vue 2 would basically have 3 choices:
1. Upgrade their codebase to Vue 3
2. Use the Webpack builder instead of Vite
3. Publish their own builder

To me, it looks like the Vue community at large is gradually maturing for Vue 3, with tools like Vuetify and Nuxt 3 soon being stable enough for adoption. This will allow most projects to migrate to Vue 3.

There will be enterprises with Vue 2 codebases that cannot be upgraded for various reasons, but that should be these enterprises' concerns. Enterprises are not resource constrained in the same way that a volunteer project is.

If any volunteers want to maintain a builder for Vue 2 (maybe some of these enterprises?), then by all means go ahead and create a separate builder!

ushironoko:
I'm sorry if I've confused you.
I appreciate the high quality work you are doing with limited resources. As you said, companies can just host plugins for Vue2 if they need to.

I "personally" wanted Vue2 support, found this PR by accident, and only saw that it may or may not be merged, and just wanted to tell you that some users simply want the feature. If it is reconsidered based on a rational decision, we will respect that decision.

PR #156, opened 2021-11-18
Only add `./` to path sometimes

### LINKED ISSUE #155: [StoryStoreV7] TypeError: importers[path] is not a function
I'm testing out 0.1.8 in my app with storyStoreV7 enabled, and it mostly works, except for one story which I have in my `.storybook` folder.  I added a console log to `importFn`, and the path being provided is `.storybook/Introduction.stories.mdx`, but the path in the `importers` object is `./.storybook/Introduction.stories.mdx`.  All the other stories seem to work fine, so I wonder if storybook is handling stories in the `.storybook` folder differently than the others, or if it's a general issue with folders starting with `.`.

IanVS:
Fixes #155 

Storybook's fileName extraction does not always add a `./` to the start of the file, only when there's not already a `.` at the start.  So this was breaking my stories inside the `.storybook/` folder.

This PR uses the same logic that storybook itself uses, so they stay in sync.

### DIFF
--- a/packages/storybook-builder-vite/codegen-importfn-script.js
+++ b/packages/storybook-builder-vite/codegen-importfn-script.js
@@ -16,7 +16,13 @@ const { normalizePath } = require('vite');
  */
 async function toImportFn(stories) {
     const objectEntries = stories
-        .map(file => `  './${normalizePath(path.relative(process.cwd(),file))}': async () => import('/@fs/${file}')`);
+        .map(file => {
+            const np = normalizePath(path.relative(process.cwd(), file));
+            // This uses the same logic as https://github.com/storybookjs/storybook/blob/056ef92e816596e077013fc261d2d1b5ff284571/lib/core-server/src/utils/StoryIndexGenerator.ts#L105,
+            // If that changes, this will need to be adjusted
+            const normalizedPath = np[0] === '.' ? np : `./${np}`;
+            return `  '${normalizedPath}': async () => import('/@fs/${file}')`
+        });
     return `
     const importers = {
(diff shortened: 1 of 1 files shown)


IanVS:
CC @tmeasday and @joshwooding would love your review on this.

tmeasday:
@IanVS you don't think this is a bug in SB itslef?

IanVS:
@tmeasday it's an inconsistency, for sure.  I'm not sure why `./` needs to be added at all, but ideally it would be for all cases.  What do you think?  I'm not sure what else is already relying on this behavior.

tmeasday:
I mean I do think it is a bug and we should fix it :). 

The idea of these paths are as JS import paths relative to the project root. So they need to start with './' or they'll be treated as imports from `node_modules`

eirslett:
> So this was breaking my stories inside the .storybook/ folder.

Do you put stories inside that folder? Sounds a bit strange to me - I usually keep the stories in the same folder as the components that they test.

joshwooding:
I think it's a bug in storybook too. Fixing it there makes the most sense to me.

IanVS:
> > So this was breaking my stories inside the .storybook/ folder.
> 
> Do you put stories inside that folder? Sounds a bit strange to me - I usually keep the stories in the same folder as the components that they test.

I put one story, a "welcome to storybook" type story in there.  It's not really past of my app, and I think it fits better in the storybook folder.

IanVS:
Closing this in favor of a fix in storybook: https://github.com/storybookjs/storybook/pull/16723

Issue #157, opened 2021-11-18
Svelte and Vue examples doesnt work


olemarius:
I cloned storybook-builder-vite (1.8.0) and tried running the example projects. React works, but Vue and Svelte fails. 

OS is Windows 11

Update:
Same thing on tag 2.0.0, but 1.3.0 works.

IanVS:
Thanks for the report.  Can you please try 1.7.0?

Also, can you share what version of storybook you are using?  And are you using features.storyStoreV7?

olemarius:
I tested 1.1.7 and 2.0.0, no changes   just ran yarn install and run sb. Will investigate more asap.

IanVS:
There's no 2.0.0, the 0.2.0 tag was a mistake.  What version of storybook are you using, and are you using storyStoreV7?

olemarius:
I'm testing again now. Apologies for being a bit unclear yesterday, had my hands a bit full.. Made sure to delete node_modules and do a clean yarn install for each tag I tested today:

0.1.3 works
0.1.7 works
0.1.8 does not work. See below screenshots.

Storybook version on 0.1.8 is Storybook 6.4.0-rc.2 for Vue3

Steps to reproduce 
1. clone https://github.com/eirslett/storybook-builder-vite.git
2. git checkout v0.1.8
3. yarn install
4. run **storybook** script in examples-vue/packages.json

![image](https://user-images.githubusercontent.com/184536/142595995-6c906033-489e-41bc-bbf3-a0ee43f5e3d3.png)
![image](https://user-images.githubusercontent.com/184536/142596301-f1a53eae-101f-4bea-ba7d-002f8fee589a.png)

IanVS:
@joshwooding any ideas what might be going on here?

olemarius:
@IanVS I had a look at https://github.com/eirslett/storybook-builder-vite/compare/v0.1.7...v0.1.8 and tried adding storyStoreV7 to v0.1.7. This triggered the same error. 

```    features: {
        storyStoreV7: true,
    },```

joshwooding:
Hmm, I think this is caused by two versions of `@storybook/client-api`

joshwooding:
This is the same issue as https://github.com/eirslett/storybook-builder-vite/issues/50

PR #160, opened 2021-11-19
Fix package duplication issues by aliasing all storybook packages

joshwooding:
Closes https://github.com/eirslett/storybook-builder-vite/issues/50
Closes https://github.com/eirslett/storybook-builder-vite/issues/157

This is the same fix as https://github.com/storybookjs/storybook/pull/11092


IanVS:
Nice fix!

PR #165, opened 2021-11-24
Fix source-loader-plugin CSF3 compatibility

### LINKED ISSUE #164: storybook-vite-source-loader-plugin + CSF3
I found an issue that works fine in Storybook `6.4.0-rc.6` + Webpack but fails with `storybook-builder-vite`.

The underlying issue is at https://github.com/eirslett/storybook-builder-vite/blob/c2e9c3cc9c7879047e793a2c441a1cdd39f576e2/packages/storybook-builder-vite/source-loader-plugin.js#L10

The `inject` function from `@storybook/source-loader` does not export a `sourceJson` property.  I don't think it ever did.  But regardless, things still generally works.

However, if I upgrade the storybook dependencies (see #161) and create a Storybook 6.4 [Component Story Format 3.0](https://storybook.js.org/blog/component-story-format-3-0/) (CSF3) test, `storybook-builder-vite` does not properly transform the story file and ends up causing Storybook to crash with a `defaultExport is undefined` message.

The following is a valid CSF3 story:

```
export default {
  title: 'Gizmo',
  render: ({ label, ...props }) => <button {...props}>{label}</button>
};

export const Default = {
  args: {
    label: 'Hello, World'
  }
};
```

The key thing to note is that there is no function export (`StoryFn`), just a CSF3 `StoryObj` export.  When transformed by `storybook-vite-source-loader-plugin` the resulting JS file is:

```
var __STORY__ = void 0;
var __LOCATIONS_MAP__ = {};
;
export const __namedExportsOrder = [];
``` 

I believe this has to do with the plugin bypassing the ["standard" Storybook transformer](https://github.com/storybookjs/storybook/blob/282e716e518b332325be97fd691807d2f76fc250/lib/source-loader/src/build.js#L3) and trying to do its own thing.

Oddly enough, if you add a standard function export (`StoryFn`), things actually work, including the CSF3 (`StoryObj`) export.

```
...
export const Weird = () => <div>Don't know why this "fixes" it.</div>;
```

joshwooding:
Closes #164

### DIFF
--- a/packages/storybook-builder-vite/source-loader-plugin.js
+++ b/packages/storybook-builder-vite/source-loader-plugin.js
@@ -1,24 +1,18 @@
-const inject =
-    require('@storybook/source-loader/dist/cjs/abstract-syntax-tree/inject-decorator').default;
+const sourceLoaderTransform =
+    require('@storybook/source-loader').default;
 
 module.exports.sourceLoaderPlugin = function () {
     return {
         name: 'storybook-vite-source-loader-plugin',
         enforce: 'pre',
-        transform(src, id) {
+        async transform(src, id) {
             if (id.match(/\.stories\.[jt]sx?$/)) {
-                const { source, sourceJson, addsMap } = inject(src, id);
-                const preamble = `
-                    /* eslint-disable */
-                    // @ts-nocheck
-                    // @ts-ignore
-                    var __STORY__ = ${sourceJson};
-                    // @ts-ignore
-                    var __LOCATIONS_MAP__ = ${JSON.stringify(addsMap)};
-                `;
+                // We need to mock 'this' when calling transform from @storybook/source-loader
+                const mockClassLoader = { emitWarning: (message) => console.warn(message), resourcePath: id };
+                const code = await sourceLoaderTransform.call(mockClassLoader, src);
 
                 return {
-                    code: `${preamble}\n${source}`,
+                    code,
                     map: { mappings: '' },
                 };
             }


packages/storybook-builder-vite/source-loader-plugin.js
```diff
@@ -1,24 +1,18 @@
-const inject =
-    require('@storybook/source-loader/dist/cjs/abstract-syntax-tree/inject-decorator').default;
+const transform =
```
IanVS, line 1: Hmmmm, I'm not sure I love using the same name as the `transform` function below.  Would you mind naming it something else, like maybe `sourceLoaderTransform` or something?

packages/storybook-builder-vite/source-loader-plugin.js
```diff
@@ -1,24 +1,18 @@
...
-                const preamble = `
-                    /* eslint-disable */
-                    // @ts-nocheck
-                    // @ts-ignore
-                    var __STORY__ = ${sourceJson};
-                    // @ts-ignore
-                    var __LOCATIONS_MAP__ = ${JSON.stringify(addsMap)};
-                `;
+                // We need to mock 'this' when calling transform from @storybook/source-loader
+                const mockClassLoader = { emitWarning: () => {}, resourcePath: id };
```
IanVS, line 11: Could we maybe use `console.log` for `emitWarning`?  I'm not sure when/how it gets used, but it would be great not to trap/drop warnings if possible.

packages/storybook-builder-vite/source-loader-plugin.js
```diff
@@ -1,24 +1,18 @@
...
+                const code = await transform.call(mockClassLoader, src);
```
IanVS, line 12: Any idea what this line is for? https://github.com/storybookjs/storybook/blob/c67c22afa75728eb00d782e88b4d43645d3a3253/lib/source-loader/src/dependencies-lookup/readAsObject.js#L12

Do we also need to mock out `extension` or whatever `getOptions` is looking for?  My guess is "no", especially since this is working as-is.

joshwooding: I don't think we do, it's webpack specific loader stuff and if options.parser or classLoader.extension is not defined the filename is.

packages/storybook-builder-vite/source-loader-plugin.js
joshwooding: Sure, it's only used when a parser can't be determined https://github.com/storybookjs/storybook/blob/c67c22afa75728eb00d782e88b4d43645d3a3253/lib/source-loader/src/abstract-syntax-tree/inject-decorator.js#L30

IanVS (approved):
awwwwyeah let's goooooo

PR #167, opened 2021-11-24
Explore adding react-docgen

joshwooding:
I know there's some work going on in storybook core (https://github.com/eirslett/storybook-builder-vite/issues/2#issuecomment-823064989) as @shilman mentioned, but I've managed to get react-docgen working (I haven't really tested it's performance) but it might be worth exploring adding this temporarily so the feature works?

### DIFF
--- a/packages/storybook-builder-vite/package.json
+++ b/packages/storybook-builder-vite/package.json
@@ -17,6 +17,7 @@
         "@storybook/csf-tools": "^6.3.3",
         "@storybook/source-loader": "^6.3.12",
         "@vitejs/plugin-react": "^1.0.8",
+        "babel-plugin-react-docgen": "^4.2.1",
         "es-module-lexer": "^0.9.3",
         "glob": "^7.2.0",
         "glob-promise": "^4.2.0",
--- a/packages/storybook-builder-vite/vite-config.js
+++ b/packages/storybook-builder-vite/vite-config.js
@@ -58,10 +58,18 @@ module.exports.pluginConfig = function pluginConfig(options, type) {
             require('@vitejs/plugin-react')({
                 // Do not treat story files as HMR boundaries, storybook itself needs to handle them.
                 exclude: [/\.stories\.(t|j)sx?$/, /node_modules/],
+                babel: {
+                    plugins: [[
+                        require.resolve('babel-plugin-react-docgen'),
+                        {
+                            DOC_GEN_COLLECTION_NAME: 'STORYBOOK_REACT_CLASSES',
+                        },
+                    ]]
+                }
             })
         );
     }
-    
+
     if (framework === 'glimmerx') {
         const plugin = require('vite-plugin-glimmerx/index.cjs');
         plugins.push(plugin.default());
--- a/yarn.lock
+++ b/yarn.lock
@@ -13618,6 +13618,7 @@ fsevents@^1.2.7:
     "@storybook/csf-tools": ^6.3.3
     "@storybook/source-loader": ^6.3.12
     "@vitejs/plugin-react": ^1.0.8
+    babel-plugin-react-docgen: ^4.2.1
     es-module-lexer: ^0.9.3
     glob: ^7.2.0
     glob-promise: ^4.2.0


packages/storybook-builder-vite/package.json
```diff
@@ -17,9 +17,11 @@
         "@storybook/csf-tools": "^6.3.3",
         "@storybook/source-loader": "^6.3.12",
         "@vitejs/plugin-react": "^1.0.8",
+        "ast-types": "^0.14.2",
         "es-module-lexer": "^0.9.3",
         "glob": "^7.2.0",
         "glob-promise": "^4.2.0",
+        "react-docgen": "6.0.0-alpha.0",
```
joshwooding, line 24: I'm using 6.0.0-alpha.0 so it supports resolving of imports. My hope is that this means we don't have to implement react-docgen-typescript although I have a working version of that too: https://github.com/joshwooding/storybook-builder-vite/tree/explore-react-docgen-typescript

packages/storybook-builder-vite/plugins/react-docgen/actualNameHandler.js
```diff
@@ -0,0 +1,52 @@
+/*
```
joshwooding, line 1: Copied from https://github.com/storybookjs/babel-plugin-react-docgen/blob/master/src/actualNameHandler.js

packages/storybook-builder-vite/plugins/react-docgen/index.js
```diff
@@ -0,0 +1,59 @@
...
+                }
+
+                let docGenExtra = '';
+
+                docgenResults.forEach((docGenResult) => {
+                    const actualName = docGenResult.actualName;
+                    //  Escape windows paths
+                    const filePath = path.relative(process.cwd(), id).replace(/\\/g, '\\\\');
+                   docGenExtra += `
+                   ${actualName}.__docgenInfo = ${JSON.stringify(docGenResult, null, 2)};
```
joshwooding, line 43: I basically reverse engineered this from the babel injection code (https://github.com/storybookjs/babel-plugin-react-docgen/blob/2df98996a51d7f537c29b63b93cc6b7687e2baaa/src/index.js#L102)

joshwooding:
There are now two solutions, one which is relying on the babel plugin (https://github.com/eirslett/storybook-builder-vite/commit/6cbc5440595fd975b6fd557788134ef53a48bb3b) and one which emulates it (https://github.com/eirslett/storybook-builder-vite/commit/480899fdb71b9e4f85225fcb03a10de94d136bdd).

Relying on the babel plugin is definitely cleaner but we can't make it easy to use `react-docgen@6.0.0-alpha.0`, maybe an alpha of `babel-plugin-react-docgen` could use it?

Currently I've added the babel plugin as a dependency but it's always included in `@storybook/react` so maybe it's not needed?

IanVS:
@joshwooding what's your sense for how important it is to use `react-docgen@6.0.0-alpha.0`?  

>Allow resolving of imported types/proptypes

Seems kind of important, right?  I think it could be okay to use your re-implementation for the time being, perhaps with a note to swap it out once (if) there's ever a stable release of react-docgen@6 (tracking issue: https://github.com/reactjs/react-docgen/issues/502).

Alternatively, like you said we could make an alpha version of babel-plugin-react-docgen, it looks like a similar approach has been taken in the past: https://github.com/storybookjs/babel-plugin-react-docgen/issues/72

IanVS:
I talked to @shilman and others about this, and they'd be willing to consider updating babel-plugin-react-docgen to use the alpha version of react-docgen if it solves the type import issue.  There's even a good chance that it will end up being faster than the typescript version that's in use right now.  We should explore it and see if it does indeed work for typescript projects.

joshwooding:
@IanVS Cool, I was looking at this today. I think releasing an alpha of babel-plugin-react-docgen is the best option.

IanVS:
I published `@ianvs/babel-plugin-react-docgen@5.0.0-alpha.0` that we can experiment with until/unless storybook merges and releases https://github.com/storybookjs/babel-plugin-react-docgen/pull/104.

PR #175, opened 2021-12-08
Allow passing options to @vitejs/plugin-vue

menthol:
This PR allow @vitejs/plugin-vue options to be set.
Then we can now use https://github.com/vuejs/rfcs/discussions/369 with storybook.

### DIFF
--- a/packages/storybook-builder-vite/README.md
+++ b/packages/storybook-builder-vite/README.md
@@ -85,6 +85,20 @@ module.exports = {
 };
 ```
 
+### Vue Customization
+
+When using this builder with VueJs, your `.storybook/main.js` (or equivalent)
+can contain a `vueOptions` object to pass custom options to
+[`@vitejs/plugin-vue`](https://github.com/vitejs/vite/tree/main/packages/plugin-vue#options):
+
+```javascript
+module.exports = {
+    vueOptions: {
+        refTransform: true,
+    },
+};
+```
+
 ## Note about working directory
 
 The builder will by default enable Vite's [server.fs.strict](https://vitejs.dev/config/#server-fs-strict)
--- a/packages/storybook-builder-vite/vite-config.js
+++ b/packages/storybook-builder-vite/vite-config.js
@@ -5,7 +5,7 @@ const { mdxPlugin } = require('./mdx-plugin');
 const { sourceLoaderPlugin } = require('./source-loader-plugin');
 
 module.exports.pluginConfig = function pluginConfig(options, type) {
-    const { framework, svelteOptions } = options;
+    const { framework, svelteOptions, vueOptions } = options;
     const plugins = [
         codeGeneratorPlugin(options),
         mockCoreJs(),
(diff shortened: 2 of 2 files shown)


IanVS:
I'm not positive that I love this pattern of creating a separate config option for each framework, but maybe it's fine.  To customize my react plugin I've been replacing it from the list in my `viteFinal`, which means I also get control over which version of the plugin is used.  I wonder if it would be better if we allowed passing a `frameworkPlugin` option, that would then just be added to the list of plugins used, and if not provided we'd fall back to what we're doing today.  I think that might be more flexible and we wouldn't have to add a special new option name for every framework we end up supporting.  What do you think, @menthol, @eirslett, and @joshwooding?

IanVS:
Oh, but thank you for submitting this PR, @menthol!  I just want to be sure we think about the pattern we're establishing here.

menthol:
> I wonder if it wo

You're right, I submitted  this PR because I don't find an easy way to avoid replacing an instantiated plugin.
Maybe having a way to override the plugin instantiation will be a lot more useful and cleaner.

IanVS:
@menthol would you like to try playing with that idea and see if it's maybe feasible?  It would be a big help, I think, and make it easier to customize the builder.

menthol:
> @menthol would you like to try playing with that idea and see if it's maybe feasible? It would be a big help, I think, and make it easier to customize the builder.

I will try

eirslett:
How is this solved with Storybook and Webpack? Can you customize plugins there too?

IanVS:
As far as I know storybook didn't provide any nice ways to handle that, other than `webpackFinal` and it was up to me to modify the config how I needed.  It's kind of tricky to do, though, and it might be nice to have a more friendly way to accomplish common configurations without needing all the power of `viteFinal`.

menthol:
> How is this solved with Storybook and Webpack? Can you customize plugins there too?


One of the major differences between webpack and vite is that webpack uses just resolved plugins while vite uses plugin instances. 

The options in webpack are therefore editable in `webpackFinal` while in `viteFinal` you can only replace  the whole plugin.

menthol:
Plus je creuse, plus je comprend que cette PR ne devrait pas exister.
La meilleur solution reste de remplacer manuellement le plugin quand on en a besoin.

Je pense que ça serais plus raisonnable de la fermer, et plus tard, d'ajouter une marche a suivre sur comment remplacer un plugin dans le README

menthol:
To be clear, there is no advantage to overriding `viteFinal`, and the way `webpackFinal` works is too different to expect to share anything.
Using an option like `frameworkPlugin` would not allow you to have the context and makes the use of overloaded plugins untenable.

menthol:
For now, what i do for my project: 
```js
require vue from '@vitejs/plugin-vue';
module.exports = {
 /* ... */
  async viteFinal(config, { configType }) {
    const vuePluginIndex = config.plugins.findIndex(({ name }) => name === 'vite:vue');
    config.plugins[vuePluginIndex] = vue({
      refTransform: true,
    });
    return config;
  },
```

IanVS:
In storybook 7.0 we will use the vite.config.js as a base, instead of always injecting our own vite framework plugins.  I think that's the best bet for how to handle this.  You can try out the 7.0 alphas right now, using `npx sb@next upgrade --prerelease`.  Thanks for this PR, but we're not going to go this direction for 6.5.

PR #190, opened 2021-12-29
Experiment with react-docgen

IanVS:
This is a similar experiment to https://github.com/eirslett/storybook-builder-vite/pull/167, but using [`@ianvs/babel-plugin-react-docgen@5.0.0-alpha.0`](https://www.npmjs.com/package/@ianvs/babel-plugin-react-docgen) so we can try out the 6.0.0-alpha.0 version of react-docgen.  

A few things I notice, which @ewlsh also pointed out in https://github.com/eirslett/storybook-builder-vite/issues/2#issuecomment-997671695:

- Union types are not handled correctly, and are shown as `union`
- Use of `React.FC` is not handled correctly
- Imported types now work, but they are shown with their typename, rather than expanding out to the actual value.  I'm not sure what `react-codegen-typescript` does, but having the actual type seems nicer than the type alias name.
- Function signatures are shown as `signature`, which doesn't really help much.

image

Despite these limitations, it might be worth thinking about moving forward in this direction to at least get _some_ kind of support, though.  Ideally, we'd give an option to also use something like `react-codegen-typescript`, but this seems like maybe a good start.

### DIFF
--- a/packages/example-react/stories/Button.stories.jsx
+++ b/packages/example-react/stories/Button.stories.jsx
@@ -4,6 +4,7 @@ export default {
     title: 'Example/Button',
     component: Button,
     argTypes: {
+        onClick: { action: 'clicked' },
         backgroundColor: { control: 'color' },
     },
 };
--- a/packages/example-react/stories/ButtonTypescript.stories.tsx
+++ b/packages/example-react/stories/ButtonTypescript.stories.tsx
@@ -1,0 +1,35 @@
+import { Button } from './ButtonTypescript';
+
+export default {
+    title: 'Example/ButtonTypescript',
+    component: Button,
+    argTypes: {
+        onClick: { action: 'clicked' },
+        backgroundColor: { control: 'color' },
+    },
+};
+
+const Template = (args) => <Button {...args} />;
+
+export const Primary = Template.bind({});
+Primary.args = {
+    primary: true,
+    label: 'Button',
+};
+
+export const Secondary = Template.bind({});
+Secondary.args = {
+    label: 'Button',
+};
+
+export const Large = Template.bind({});
+Large.args = {
+    size: 'large',
+    label: 'Button',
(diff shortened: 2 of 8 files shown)


ewlsh:
I added some context on our Vite plugin work to #2, I'd be curious if custom handlers receive a TypeScript AST in this implementation.

joshwooding:
I think my thoughts were to provide at least some support for now. 
We might want to add some custom handlers, that we can push upstream so everyone benefits but I don't think we should necessarily worry about making sure everything is fully supported before we release.

It's definitely possible to get `react-docgen-typescript` working and we might want to support it after this basic support. I know that I'll need it for my storybooks

IanVS:
We now have support for react-docgen-typescript thanks to @joshwooding.  I think it's maybe worth also supporting this option, for those not using typescript.  If there's interest, I'll clean this up and get it ready for review/merge.

reintroducing:
@IanVS non-TS user here weighing in from the peanut gallery. I definitely think this should be supported for those that choose not to use TS, I think that's a pretty common use case and all these tools support both TS and non-TS projects in them.

IanVS:
Thanks @reintroducing.  I'll work on this a bit.  We may find a better way in the future, but it would be good to support folks for now, and this is pretty close to being ready I think.

sebastienbarre:
Thanks @IanVS for your work -- definitely a feature I'm missing as well after switching from Webpack to the Vite plugin.

IanVS:
@reintroducing and @sebastienbarre thanks for the nudges.  I've opened https://github.com/storybookjs/builder-vite/pull/299 which I think is a better approach to add support for react-docgen, if you'd like to keep an eye on that one.

reintroducing:
@IanVS thank you! i've gone ahead and subscribed to that PR, appreciate your work on this.

Issue #188, opened 2021-12-26
Suggestions for improving maintainability


mrauhu:
Hello.

I want to implement best practices in tooling that inspired by [Storybook][s] and [Vite][v] repositories.

[s]: https://github.com/storybookjs/storybook/
[v]: https://github.com/vitejs/vite

The goal is to make it easier for maintainers to participate in the project.

Related:  #11

### Zero changes to the `storybook-builder-vite` package code

- [ ] Move the Prettier configuration from `package.json` to `.prettierrc`.
- [ ] Synchronize `.editorconfig` with the Prettier standard.
- [ ] Add ESLint with Prettier plugin.
- [ ] Add Jest for unit testing.
- [ ] Add Cypress for E2E testing.
- [ ] Running tests on a pull request and publish events.

### Minor changes to the package code: syntax and typing

- [ ] Run `eslint --fix` on all codebase.
- [ ] Use TypeScript (with right `tsconfig.json` for comfort debugging).
- [ ] Add the TypeScript plugin to ESlint.
- [ ] Convert all codebase from `*.js` to `*.ts`.

### Minor changes in examples `package.json` and Github Actions

- [ ] Workspaces: use NPM or Pnpm, because Yarn is not working out of the box like it should be.

@eirslett, @IanVS 

May I create a one pull request with all suggested features?

Best wishes,
Sergey.

IanVS:
Thanks for the suggestions, @mrauhu, and for checking before submitting a PR.  I think it's good to talk about this kind of thing before putting in the work.  

I can't speak for @eirslett, and this is his project, but I'm personally on board with most of your suggestions.  One area I disagree in though is the use of TypeScript.  Normally I am a big fan of TypeScript, and I tend to use it in my projects.  However, the experimental nature of this builder means that I'm often mucking around with it inside my node_modules of the application I'm building, and many times I've appreciated that the source is not compiled, and I can work directly with the code rather than a minified or compiled output.  For that reason, at least for now, I'd recommend sticking with plain old javascript without a build step.  

As for switching to a different package manager, yarn is used by Storybook itself, so if it's falling down on the job we're probably doing something wrong.  Can you explain what you mean by not working as it should?  

For eslint & prettier, there are different ways to get them to work together.  One is to run prettier through eslint, and the other is to disable all the eslint style rules, and then use prettier to do formatting and eslint for other linting.  Which approach do you propose?  

[...]

PR #193, opened 2022-01-05
Improving maintainability: ESLint and Prettier

mrauhu:
Hello.

This pull request is part of https://github.com/eirslett/storybook-builder-vite/issues/188.

Done:

- [x] Move the Prettier configuration from `package.json` to `.prettierrc`.
- [x] Synchronize `.editorconfig` with the Prettier standard.
- [x] Add ESLint with Prettier config.
- [x] Add the TypeScript plugin to ESlint.
- [x] Run Prettier before ESLint 

@eirslett, @IanVS

Best wishes,
Sergey.


package.json
```diff
@@ -1,25 +1,29 @@
...
+  "name": "storybook-builder-vite-parent",
+  "version": "0.1.13",
+  "description": "An experimental plugin to run and build Storybooks with Vite",
+  "license": "MIT",
+  "engines": {
+    "node": ">=16.0.0"
+  },
+  "scripts": {
+    "lint": "cross-env NODE_ENV=production eslint \"packages/*/**/*.{js,jsx,mjs,cjs}\" --fix",
+    "lint-ci": "cross-env NODE_ENV=production eslint \"packages/*/**/*.{js,jsx,mjs,cjs}\""
```
joshwooding, line 11: We shouldn't need to use `cross-env` here since we are using yarn berry.

joshwooding (dismissed):

package.json
```diff
@@ -1,25 +1,29 @@
...
-    }
+  "name": "storybook-builder-vite-parent",
+  "version": "0.1.13",
+  "description": "An experimental plugin to run and build Storybooks with Vite",
+  "license": "MIT",
+  "engines": {
+    "node": ">=16.0.0"
+  },
+  "scripts": {
+    "lint": "cross-env NODE_ENV=production eslint \"packages/*/**/*.{js,jsx,mjs,cjs}\" --fix",
```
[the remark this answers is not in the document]
mrauhu: > If I understand why this is needed we don’t need `NODE_ENV=production` here.

Because, we have a few rules for that disabled for development in ESLint config, see:

https://github.com/eirslett/storybook-builder-vite/blob/34a690bcb6e2ac1f7a8810b47e748574ca5434ca/.eslintrc.js#L14-L16

So, when you use IDE with ESLint plugin, you will not get warnings about using `console` and `debugger`.

But before committing and publishing we need to run linting with enabled rules, for manual code cleanup from breakpoints and and debug output.

package.json
mrauhu: Part of the plan https://github.com/eirslett/storybook-builder-vite/issues/188 is:

> Workspaces: use NPM or Pnpm, because Yarn is not working out of the box like it should be.

More info about it in https://github.com/eirslett/storybook-builder-vite/issues/188#issuecomment-1004221826.

So, I used a cross-platform solution for scripts that works independently of the package manager.

IanVS: Maybe we can wait until that PR before making the change, if it's needed there?

package.json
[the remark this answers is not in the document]
IanVS: I think it should be up to developers to configure their editors to ignore those warnings, if that's what they would like.

.editorconfig
```diff
@@ -1,8 +1,12 @@
 root=true
 
 [*]
-indent_style = space
-indent_size = 4
 charset = utf-8
+max_line_length = 120
```
IanVS, line 5: The default line-length for prettier is 80, and it isn't being overridden in the .prettierrc, so I guess it should be 80 here as well, right?

package.json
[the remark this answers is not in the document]
joshwooding: Hmm, I saw this but I've never come across a situation where I've wanted to built this into eslint's config.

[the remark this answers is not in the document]
mrauhu: > I think it should be up to developers to configure their editors to those warnings, if that's what they would like.

@IanVS It's already configured in ESLint rules.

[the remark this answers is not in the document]
mrauhu: > my confusion is that we have lint and lint:ci both are production even though to me lint would be development?

@joshwooding  

* `lint` — used before committing changes, have `--fix` argument;
* `lint-ci` — used before publishing.

They are in production mode, because we need show warning if developer is not deleted `console.*` methods or `debugger` statements.

package.json
mrauhu: > Maybe we can wait until that PR before making the change, if it's needed there?

Yes, I can remove the `cross-env`.

package.json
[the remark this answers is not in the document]
IanVS: Right, but I think it shouldn't be.  I think they should be warnings no matter the `NODE_ENV`.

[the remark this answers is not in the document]
joshwooding: I think there should always be warnings too. This could cause a lot of confusion.

.nvmrc
```diff
@@ -1 +0,0 @@
-v15.14.0
```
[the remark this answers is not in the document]
mrauhu: As far I know the `.nvmrc` file is part of the `nvm` package.

There is a native `package.json` field called the `engines`:

https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines

I'm already using it:

https://github.com/eirslett/storybook-builder-vite/blob/34a690bcb6e2ac1f7a8810b47e748574ca5434ca/package.json#L6-L8

.editorconfig
mrauhu: @IanVS no, we are using the Prettier from ESLint:

https://github.com/eirslett/storybook-builder-vite/blob/34a690bcb6e2ac1f7a8810b47e748574ca5434ca/.eslintrc.js#L12-L13

package.json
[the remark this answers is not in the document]
mrauhu: > I think there should always be warnings too.

Instead of focusing on real warnings and errors, you got false positive warnings about `console` and `debugger`:

image

It's very annoying when you writing a code.

So, my solution is provide a way for a developer to focus on the real issues, not a ghosts:

image

But if I run the `lint` script I can catch the false-positive warning before committing the code:

image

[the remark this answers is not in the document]
mrauhu: @IanVS @joshwooding how you want to setup `no-console` and `no-debugger` ESLint rules for better developer experience:

* hide false-positive warnings in the development mode, show them in the production mode;
* show them always;
* disable this rules?

package.json
mrauhu: Fixed in c8e135c.

package.json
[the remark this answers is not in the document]
mrauhu: Resolved in 44fce07.

.editorconfig
mrauhu: Updated `.prettierrc` in 12279cd.

.eslintrc.js
```diff
@@ -0,0 +1,28 @@
...
+    node: true,
+  },
+  parserOptions: {
+    ecmaVersion: 2021,
+  },
+  plugins: ['only-warn'],
+  extends: ['eslint:recommended', 'plugin:prettier/recommended', 'plugin:storybook/recommended'],
+  rules: {
+    // Set line length to 120
+    'max-len': ['error', { code: 120 }],
```
IanVS, line 13: If prettier is controlling the line-length, why does eslint also need to set the same value?  Shouldn't formatting be left entirely up to prettier?

.eslintrc.js
```diff
@@ -0,0 +1,28 @@
+module.exports = {
+  root: true,
+  env: {
+    node: true,
+  },
+  parserOptions: {
+    ecmaVersion: 2021,
+  },
+  plugins: ['only-warn'],
```
IanVS, line 9: Can you explain the reason for only using warnings?  These tend to not show up well in github actions, and it's easy to miss them and accidentally merge a PR with warnings, only to notice them later on in a different PR.

package.json
```diff
@@ -1,25 +1,28 @@
...
+  "name": "storybook-builder-vite-parent",
+  "version": "0.1.13",
+  "description": "An experimental plugin to run and build Storybooks with Vite",
+  "license": "MIT",
+  "engines": {
+    "node": ">=16.0.0"
+  },
+  "scripts": {
+    "lint": "eslint \"packages/*/**/*.{js,jsx,mjs,cjs}\" --fix",
+    "lint-ci": "eslint \"packages/*/**/*.{js,jsx,mjs,cjs}\""
```
IanVS, line 11: I think we still need a scripts to run prettier to check that the formatting is correct.  I see you have eslint set up to disable the style rules using https://github.com/prettier/eslint-config-prettier, but aside from prettier running in editors, there's nothing that will enforce a correct style in CI, as far as I can tell.

.eslintrc.js
mrauhu: It's simple, we reduce count of vertical lines in IDEs that is created by different plugins (EditorConfig, ESLint, Prettier) and set it all to same value.

## Example

### Good

Single line:

image

### Bad

Multiple lines:

image

package.json
mrauhu: We have ESLint, It works in IDEs or Terminal/Command line.

Can you please explain, why you need the Prettier for check formatting?

> enforce a correct style in CI

I think, you doing something wrong, because in CI stage you only check code.

Can you provide an example of enforcing a correct style in CI?

Also, right now we have zero style checks in CI (GIthub Actions).

.eslintrc.js
mrauhu: I don't like errors in IDE from linter, because it's very hard to understand what is not working, it's bad practice that results to missing the real errors.

If you want getting errors in CI, I can fix it using [solution](https://github.com/bfanger/eslint-plugin-only-warn#usage):

```json5
{
  "scripts": {
     "lint": "eslint --max-warnings=0 ...",
   }
}
```

package.json
IanVS: I believe the main point of using prettier is to enforce a consistent code style.  If we are not checking it in CI (which you're right, we aren't right now, but we should), then someone can commit a change without running prettier, and then someone else in another PR will run prettier and change lines that are not related to their PR.  By enforcing that each PR that is merged is correctly formatted, we can guarantee that prettier will not change unrelated lines.

.eslintrc.js
IanVS: Sorry, I don't understand, those two screenshots look like completely different import statements.

By using `plugin:prettier/recommended`, you've disabled all of eslint's style rules, so that prettier can handle them.  But then you turn back on the max-len rule, which is going to fight with prettier because they might break lines up differently.  Better to keep all style formatting within prettier, and let eslint only warn about syntax and best-practices.

package.json
mrauhu: We can avoid it by:

* using ESLint before commit changes:

  ```
  yarn lint
  ```
* using ESLint in CI stage

  ```
  yarn lint-ci
  ```

I already done it.

IanVS: `yarn lint` only runs eslint.  ESLint is configured to not enforce style rules, deferring to prettier.  How is prettier formatting done as a part of this?  The config includes `eslint-config-prettier`, which `Turns off all rules that are unnecessary or might conflict with Prettier.`

I see you added `eslint-plugin-prettier` to the devDependencies, but it is not being used in the config.  Nor do I think it should be.  I am of the opinion that prettier should be run on its own, not via eslint, as recommended in https://prettier.io/docs/en/integrating-with-linters.html.

.eslintrc.js
mrauhu: @IanVS 

> those two screenshots look like completely different import statements, and I don't see how they relate to eslint or prettier.

First screenshot from `storybook-builder-vite` (this pull request), second from `storybook` (`next` branch). 

Plugins of a JetBrains IDE automatically getting configs from ESLint and Prettier.

So, in the Storybook we have two different values of max line lenght, one from the Prettier, and another from imported nested config of the ESLint.

But in this pull request I set the same value for all plugins (EditorConfig, ESLint and Prettier), so you see the one vertical line.

> By using plugin:prettier/recommended, you've disabled all of eslint's style rules, so that prettier can handle them.

Yes, I do.

> But then you turn back on the max-len rule, which is going to fight with prettier because they might break lines up differently.

I don't think so, because I set the same value — `120`.

> Better to keep all style formatting within prettier, and let eslint only warn about syntax and best-practices.

Impossible, but I done it in this pull request.

Right now all done as you described:

* we keeping all style formatting within Prettier plugin;
* and warn  about syntax and best practices by using other plugins.

All in one command:

```
eslint
```

package.json
mrauhu: > I see you added `eslint-plugin-prettier` to the devDependencies, but it is not being used in the config.

You are wrong.

I do it in `extends`:

https://github.com/eirslett/storybook-builder-vite/blob/22c3b98c9b9bb935cfb8d76de8d306ce0cc21607/.eslintrc.js#L10

.eslintrc.js
mrauhu: Fixed in 49ac41cc.

.nvmrc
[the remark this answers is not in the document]
eirslett: That's right, but nvm ignores `package.json`. Ideally it should read whatever is specified there, but instead we have to duplicate it into the `.nvmrc` file.

.editorconfig
eirslett: Actually it's not necessary to reflect this setting; prettier reads `.editorconfig` files and respects them. Pretty convenient. https://github.com/prettier/prettier/blob/514046b3c70e6477cb69b4d36871d0d4c8c5e415/src/config/resolve-config-editorconfig.js

package.json
eirslett: We could do something like
```
"lint-ci": "eslint \"packages/*/**/*.{js,jsx,mjs,cjs}\" && prettier --check ."
``` 

which would first run eslint, and then prettier. But it could be redundant (time-wasting) - the eslint preset for prettier is supposed to run pretty much the exact same rules that `prettier --check` does. There could be discrepancies due to differences in implementation between eslint and prettier, but that's more of a theoretical concern, I believe. 🤷

.editorconfig
mrauhu: @eirslett Okay, I can remove setting from `.prettierrc`.

But I will keep it in `.eslintrc`, because we need to multiple vertical lines in IDEs.

package.json
mrauhu: > We could do something like

@eirslett it's already done by:

https://github.com/eirslett/storybook-builder-vite/blob/49ac41cccf5dbd74be99c761bf2afad7d0a659ee/package.json#L11

> But it could be redundant (time-wasting) - the eslint preset for prettier is supposed to run pretty much the exact same rules that prettier --check does.

Yes, you are right.

mrauhu:
I published a small portion of changes, and I got a lot of questions.

Spoilers:

Configuration of ESLint with Prettier plugins is based on the best practices from the Vue and Vite community.

.editorconfig
eirslett: > But I will keep it in .eslintrc, because we need to multiple vertical lines in IDEs.

Or maybe use something like this - I've never tried it myself though. https://www.npmjs.com/package/eslint-plugin-editorconfig

mrauhu: Fixed in 4ce1418.

eirslett:
It's looking very good so far!

Code style PRs are always a PITA since everybody has so strong opinions about code style! I couldn't care less if it's 2 or 4 spaces or tabs or whatever, 80 characters, 120 characters, I just really like going with the default/recommended settings on everything. That's what I always do on greenfield projects.

The trade-off in this case is for example the indent size, which will lead to merge conflicts in the other 10 open PRs, and git blame will be a bit less useful unless you ignore whitespace diffs (but only "power users" of the git tools do that, in my experience).

mrauhu:
> The trade-off in this case is for example the indent size, which will lead to merge conflicts in the other 10 open PRs

@eirslett 

Problem is, right now there is no code style, because in every file you can find different indent size.

After this pull-request is merged, you need do once for each pull request to run:

```
yarn lint
```

Then commit and push changes.

.nvmrc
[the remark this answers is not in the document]
mrauhu: >  nvm ignores `package.json`.

Fixed in aed50085.

IanVS:
I checked out the branch, ran `yarn lint`, and got a number of errors and warnings, including errors like this:

```
/Users/ianvs/code/storybook/storybook-builder-vite/packages/storybook-builder-vite/index.js
  25:1  warning  This line has a length of 124. Maximum allowed is 120  max-len
```

I believe these are due to long comments and strings which prettier does not break into multiple lines.

There are also a number of parsing errors in .jsx files, like this:

```
/Users/ianvs/code/storybook/storybook-builder-vite/packages/example-workspaces/packages/app/stories/Button.jsx
  12:9  error  Parsing error: Unexpected token <
```

Lastly, there are a number of files which prettier wants to format but which are not being formatted by the current `yarn lint` command.  This can cause a problem if maintainers have prettier enabled in their editors, since formatting changes will be made for unrelated lines whenever anyone works in those files.

```
▶ yarn prettier --check .
Checking formatting...
[warn] packages/example-react/.storybook/main.js
[warn] packages/example-react/package.json
[warn] packages/example-react/stories/button.css
[warn] packages/example-react/stories/Button.jsx
[warn] packages/example-react/stories/Button.stories.jsx
[warn] packages/example-react/stories/EnvironmentVariables.jsx
[warn] packages/example-react/stories/EnvironmentVariables.stories.jsx
[warn] packages/example-react/stories/header.css
[warn] packages/example-react/stories/Header.jsx
[warn] packages/example-react/stories/Header.stories.jsx
[warn] packages/example-react/stories/Introduction.stories.mdx
[warn] packages/example-react/stories/mdx-in-stories/Example.docs.mdx
[warn] packages/example-react/stories/mdx-in-stories/Example.stories.jsx
[warn] packages/example-react/stories/page.css
[warn] packages/example-react/stories/Page.jsx
[warn] packages/example-react/stories/Page.stories.jsx
[warn] packages/example-svelte/.storybook/main.js
[warn] packages/example-svelte/.storybook/preview.js
[warn] packages/example-svelte/package.json
[warn] packages/example-svelte/stories/button.css
[warn] packages/example-svelte/stories/header.css
[warn] packages/example-svelte/stories/Introduction.stories.mdx
[warn] packages/example-svelte/stories/page.css
[warn] packages/example-vue/.storybook/main.js
[warn] packages/example-vue/.storybook/preview.js
[warn] packages/example-vue/package.json
[warn] packages/example-vue/stories/button.css
[warn] packages/example-vue/stories/Button.vue
[warn] packages/example-vue/stories/EnvironmentVariables.vue
[warn] packages/example-vue/stories/header.css
[warn] packages/example-vue/stories/Header.vue
[warn] packages/example-vue/stories/Introduction.stories.mdx
[warn] packages/example-vue/stories/page.css
[warn] packages/example-vue/stories/Page.vue
[warn] packages/example-workspaces/package.json
[warn] packages/example-workspaces/packages/app/stories/button.css
[warn] packages/example-workspaces/packages/app/stories/Button.jsx
[warn] packages/example-workspaces/packages/app/stories/Button.stories.jsx
[warn] packages/example-workspaces/packages/app/stories/EnvironmentVariables.jsx
[warn] packages/example-workspaces/packages/app/stories/EnvironmentVariables.stories.jsx
[warn] packages/example-workspaces/packages/app/stories/header.css
[warn] packages/example-workspaces/packages/app/stories/Header.jsx
[warn] packages/example-workspaces/packages/app/stories/Header.stories.jsx
[warn] packages/example-workspaces/packages/app/stories/Introduction.stories.mdx
[warn] packages/example-workspaces/packages/app/stories/mdx-in-stories/Example.docs.mdx
[warn] packages/example-workspaces/packages/app/stories/mdx-in-stories/Example.stories.jsx
[warn] packages/example-workspaces/packages/app/stories/page.css
[warn] packages/example-workspaces/packages/app/stories/Page.jsx
[warn] packages/example-workspaces/packages/app/stories/Page.stories.jsx
[warn] packages/example-workspaces/packages/catalog/.storybook/main.js
[warn] packages/example-workspaces/packages/catalog/package.json
[warn] packages/storybook-builder-vite/input/iframe.html
[warn] packages/storybook-builder-vite/package.json
[warn] packages/storybook-builder-vite/README.md
[warn] README.md
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
```

IanVS:
>git blame will be a bit less useful unless you ignore whitespace diffs

It's not just whitespace changes which will be made when we run formatting.  I appreciate that the formatting is not being done in this PR, because if we do all of the formatting in a separate PR, we can use [git-ignore-revs-fie](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt) to hide the formatting commit from git blame.

package.json
IanVS: > the eslint preset for prettier is supposed to run pretty much the exact same rules that prettier --check does.

Unfortunately this doesn't seem to be the case, see https://github.com/eirslett/storybook-builder-vite/pull/193#issuecomment-1006205181.

mrauhu:
> I checked out the branch, ran yarn lint, and got a number of errors and warnings ...

Yes, because all formatting is done in #195.

mrauhu:
> This can cause a problem if maintainers have prettier enabled in their editors

@IanVS I'm already working on the `CONTRIBUTING.md`, so there will be no problems.

joshwooding:
> > This can cause a problem if maintainers have prettier enabled in their editors
> 
> 
> 
> @IanVS I'm already working on the `CONTRIBUTING.md`, so there will be no problems.

I'm just curious why we wouldn't want our IDEs to run prettier?

package.json
```diff
@@ -1,25 +1,28 @@
...
-    }
+  "name": "storybook-builder-vite-parent",
+  "version": "0.1.13",
+  "description": "An experimental plugin to run and build Storybooks with Vite",
+  "license": "MIT",
+  "engines": {
+    "node": ">=16.0.0"
+  },
+  "scripts": {
+    "lint": "eslint \"packages/*/**/*.{js,jsx,mjs,cjs}\" --fix",
```
IanVS, line 10: ESLint is not currently configured correctly for `.jsx` files, let's remove it from this list.  And actually, I think as a starting point it would be good to run eslint only on `packages/storybook-builder-vite`, and the examples should probably have their own linting configs, since react will be different from vue and from svelte, etc.

mrauhu:
> I'm just curious why we wouldn't want our IDEs to run prettier? It's an easy workflow which almost always ensures formatting is kept consistent.

It's simple:

> Linters usually contain not only code quality rules, but also stylistic rules. Most stylistic rules are unnecessary when using Prettier, but worse – they might conflict with Prettier! 
> ...
> Luckily it’s easy to turn off rules that conflict or are unnecessary with Prettier, by using these pre-made configs:
> 
> `eslint-config-prettier`

https://prettier.io/docs/en/integrating-with-linters.html

Easy workflow which almost always ensures formatting is kept consistent:

image

package.json
mrauhu: Right now I have zero ESLint warnings and errors in #195 pull request.

So, for me It's works for all codebase (also for `*.jsx`).

If you want to provide additional configs, let's do it after this and #195 pull requests will be merged.

joshwooding:
> > I'm just curious why we wouldn't want our IDEs to run prettier? It's an easy workflow which almost always ensures formatting is kept consistent.
> 
> 
> 
> It's simple:
> 
> 
> 
> > Linters usually contain not only code quality rules, but also stylistic rules. Most stylistic rules are unnecessary when using Prettier, but worse – they might conflict with Prettier! 
> 
> > ...
> 
> > Luckily it’s easy to turn off rules that conflict or are unnecessary with Prettier, by using these pre-made configs:
> 
> > 
> 
> > `eslint-config-prettier`
> 
> 
> 
> https://prettier.io/docs/en/integrating-with-linters.html
> 
> 
> 
> Easy workflow which almost always ensures formatting is kept consistent:
> 
> 
> 
> image
> 
> 

If we want to start quoting the integrating with linters page we can't skip over: 

image

Which mentions that eslint-plugin-prettier is not generally recommended with a list of downsides :)

IanVS:
@mrauhu Yes, you're using the config that disables eslint style rules, which is great.  But eslint does not run on every file that prettier can run on.  I think that's the main challenge here.  I'm still unclear on why you prefer to run prettier via eslint.  I haven't heard any reasons other than mentioning it's done in vue and vite, but maybe they have different needs.

The downsides of this approach are:

- Formatting is only applied to files that eslint is run on, not `.css`, `.md`, etc.
- It's an extra eslint plugin that we need to use.
- Prettier itself discourages this approach by saying:

>These plugins were especially useful when Prettier was new. By running Prettier inside your linters, you didn’t have to set up any new infrastructure and you could re-use your editor integrations for the linters. But these days you can run prettier --check . and most editors have Prettier support.
>
>The downsides of those plugins are:
>
>- You end up with a lot of red squiggly lines in your editor, which gets annoying. Prettier is supposed to make you forget about formatting – and not be in your face about it!
>- They are slower than running Prettier directly.
>- They’re yet one layer of indirection where things may break.

mrauhu:
I made all suggested changes.

Please, merge it.

package.json
mrauhu: We can add other plugins later.

Current set of plugins is working great for the `storybook-builder-vite` package.

IanVS:
>Please, merge it.

I appreciate all the hard work you're putting into this, @mrauhu, and I hope that I'm not coming off as too demanding or nit-picky.  But I would like to be sure that we are not causing trouble for ourselves down the line.  @joshwooding and I both work in this repo fairly often, and we have experience working in other projects as well.  In fact, I was one of the early maintainers of ESLint back around its 1.0 timeframe.  So I think we are not speaking out of ignorance.

I'm going to leave the final say here to @eirslett, since this is his repo.  But here are my main concerns:

- I think PRs should stand alone and contain a complete changeset.  As it is, if we merge this PR without #195, we will have a broken eslint config.  
  - *What I would like to see*: This PR should contain all the necessary config such that a small subsequent PR should only need to fix lint warnings and formatting (not config).  The commit SHA of that subsequent PR can then be added to a file and we can use `--ignore-revs-file` to exclude those changes from git blame.
- I have a strong preference for running prettier apart from eslint, for all the reasons I and @joshwooding mentioned above.  I haven't heard any good reasons here for tying prettier into eslint.  
  - *What I would like to see*: eslint-plugin-prettier should be removed, scripts should be added for `prettier --write` and `prettier --check`.  
- The eslint config should be the same no matter the `NODE_ENV`.  If you don't want to see certain errors in your editor, you should configure your editor thusly.  
  - *What I would like to see*: Remove the rules for 'no-console' and 'no-debugger' from the eslint config

There are other decisions here which I don't agree with, but I have expressed them and don't care enough about them to press those points.  But without the changes I've listed above, I won't be approving this PR, but I won't block it either, since like I said, I defer to @eirslett.

Thanks again for all your work here, and your efforts to improve the codebase.

mrauhu:
@IanVS thank you for clarification.

IanVS (approved):
@mrauhu thank you for understanding and for making the changes I requested.  I think this is in a great spot now.

eirslett:
> if we merge this PR without Improving maintainability: TypeScript #195, we will have a broken eslint config.

That's fine, we can live with a somewhat broken eslint config as long as it is temporary (for a few hours or days).

Are there any blockers to merging this PR now, as it is? It looks complete to me.

The only thing is that there are 28 commits (as of now) and many of these commits are simple reverts of one another, so the history is confusing. @mrauhu could you please squash all the changes into 1 commit, with a good commit message? Then we'll get this stuff merged!

mrauhu:
@eirslett 

> > if we merge this PR without Improving maintainability: TypeScript #195, we will have a broken eslint config.
> 
> That's fine, we can live with a somewhat broken eslint config as long as it is temporary (for a few hours or days).

Right now there is nothing broken, I moved the ESLint config changes from #195 to #193.

Aslo, I reorder commits between three pull requests:

1. #193 — this pull request;
2. #196 — on top of 193, ran `lint` and fixed warnings in the package directory;
3. #195 — on top of 196, converted codebase from JavaScript to TypeScript.

I'm working on typization for #195 draft and soon convert it to pull request.

> ... could you please squash all the changes into 1 commit, with a good commit message?

I can, but these pull requests are on top of each other, can you merge them one by one without squashing?

mrauhu:
@eirslett about squashing:

You can enable the Allow squash merging  option in the repo settings:

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests

mrauhu:
> The only thing is that there are 28 commits (as of now) and many of these commits are simple reverts of one another, so the history is confusing.

I think history of commits provide a better understanding for a developer what is changed and why.

It works great with an IDE, because you can view commit history for every file and line.

eirslett:
Ok, there we go! Thanks a lot for the work! 🥳

PR #198, opened 2022-01-07
Improving maintainability: fix ignore revs

mrauhu:
Hello.

Updated a commit id in the `.git-blame-ignore-revs` file, because commit id was changed after squash merge.

@eirslett, @IanVS 

Please, do not use the squash merge method, because it's:

* breaks commit IDs, as result the `.git-blame-ignore-revs` file have wrong IDs;
* breaks ability to merge branches with same commits (on top of each other), so we are need to manually rebase inherit branches after every squash merge.

Best wishes,
Sergey.


joshwooding:
Thanks @mrauhu

eirslett:
The squash merge was done on purpose. Please respect this decision.

PR #216, opened 2022-01-24
Fix params.fileName being null when storyStoreV7 is false

### LINKED ISSUE #142: parameters.fileName is null
hello.
As you can see in the screenshot, `parameters.fileName` is null.
Add-ons that depend on'fileName' won't work.
(e.g. storybook-addon-playwright )

image

story source
```
const Template: ComponentStory<typeof Button> = function (args) {
  console.log(arguments);
  return <Button {...args} />;
};
```

I have patched `codegen-iframe-script.js` as follows.

add this code in line 56.
```
        const cwd = process.cwd();
        const resolveMap = storyEntries.reduce((prev, entry) => ({...prev, [entry]: entry.replace(cwd, '.') }), {});
        const modules = storyEntries.map((entry, i) => `${JSON.stringify(entry)}: story_${i}`).join(',')
        const loadable = `
        (() => {
          const loadable = function(key) {
            return {${modules}}[key];
          };
          Object.assign(loadable, {
            keys: () => (${JSON.stringify(Object.keys(resolveMap))}),
            resolve: (key) => (${JSON.stringify(resolveMap)}[key])
          });
          return loadable;
        })()`;
```

change line 112.
```
    configure(${loadable}, { hot: import.meta.hot }, false); // not sure if the import.meta.hot thing is correct
    `.trim();
        return code;
    };
```

joshwooding:
Closes #142

Thank you @wild-nyankoman for the solution. I made some minor tweaks to styling and added support for windows.

### DIFF
--- a/packages/storybook-builder-vite/codegen-iframe-script.ts
+++ b/packages/storybook-builder-vite/codegen-iframe-script.ts
@@ -1,7 +1,7 @@
 import { loadPreviewOrConfigFile } from '@storybook/core-common';
 import { normalizePath } from 'vite';
 import { listStories } from './list-stories';
-import { default as slash } from 'slash';
+import slash from 'slash';
 
 import type { ExtendedOptions } from './types';
 
@@ -25,7 +25,7 @@ export async function generateIframeScriptCode(options: ExtendedOptions) {
   const configEntries = [...presetEntries, previewOrConfigFile].filter(Boolean);
 
   const storyEntries = await listStories(options);
-  const resolveMap = storyEntries.reduce(
+  const resolveMap = storyEntries.reduce<Record<string, string>>(
     (prev, entry) => ({ ...prev, [entry]: entry.replace(slash(process.cwd()), '.') }),
     {}
   );
--- a/packages/storybook-builder-vite/codegen-iframe-script.ts
+++ b/packages/storybook-builder-vite/codegen-iframe-script.ts
@@ -1,6 +1,7 @@
 import { loadPreviewOrConfigFile } from '@storybook/core-common';
 import { normalizePath } from 'vite';
 import { listStories } from './list-stories';
+import { default as slash } from 'slash';
 
 import type { ExtendedOptions } from './types';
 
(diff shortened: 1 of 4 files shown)


packages/storybook-builder-vite/codegen-iframe-script.ts
```diff
@@ -1,6 +1,7 @@
 import { loadPreviewOrConfigFile } from '@storybook/core-common';
 import { normalizePath } from 'vite';
 import { listStories } from './list-stories';
+import { default as slash } from 'slash';
```
IanVS, line 4: Question: why not just `import slash from 'slash'`?

packages/storybook-builder-vite/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "strict": true,
     "rootDir": ".",
-    "outDir": "dist"
+    "outDir": "dist",
+    "esModuleInterop": true
```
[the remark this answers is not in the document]
IanVS: If we're adding this to our own dependencies, is there any reason we couldn't use 4.0 right now?

packages/storybook-builder-vite/codegen-iframe-script.ts
```diff
@@ -24,6 +25,11 @@ export async function generateIframeScriptCode(options: ExtendedOptions) {
   const configEntries = [...presetEntries, previewOrConfigFile].filter(Boolean);
 
   const storyEntries = await listStories(options);
+  const resolveMap = storyEntries.reduce(
```
IanVS, line 28: Without specifying a generic for `reduce`, typescript will infer the type of `resolveMap` based on the second argument, which is an empty object.  Since we know the result of the reduce is an object with string keys and string values, we can use this:

```suggestion
  const resolveMap = storyEntries.reduce<Record<string, string>>(
```

packages/storybook-builder-vite/tsconfig.json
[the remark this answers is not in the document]
joshwooding: Since the package is pure ESM it means we can't `require()` it. We output CommonJS so it breaks :(

packages/storybook-builder-vite/codegen-iframe-script.ts
joshwooding: Hmm, I was seeing a weird error but I think that was in a different setup. I can just use the normal import

IanVS (approved):
LGTM!

PR #217, opened 2022-01-24
Remove unfetch import

joshwooding:
Copying [#17306](https://github.com/storybookjs/storybook/pull/17306). I thought it was unneccessary but, now it's been removed we can safely remove it too.


IanVS (approved):
Nice!

PR #218, opened 2022-01-24
Enable auto-title for non-storyStoreV7

### LINKED ISSUE #201: Autotitle when not using storyStoreV7
Currently, autotitle does not work without enabling storyStoreV7.  We need to provide a global `STORIES` like the webpack builder does here: https://github.com/storybookjs/storybook/blob/e0c0a1429dc123219dfd74c8c4a1c6bfc9660f74/lib/builder-webpack4/src/preview/iframe-webpack.config.ts#L191-L194, so that it can be used here: https://github.com/storybookjs/storybook/blob/e0c0a1429dc123219dfd74c8c4a1c6bfc9660f74/lib/client-api/src/StoryStoreFacade.ts#L156-L164

I've hit two snags so far.  One, is that autoTitle uses `path`, which webpack polyfills but vite does not (there's a simple solution I have working, and I'll submit a PR to storybook soon).  Second, the format of the loadable function we are providing to the call to `configure` does not include the filenames, which means that autoTitle does not name the stories correctly, defaulting to `export_map_1`, rather than the actual filename.  Here is where the loadable is used: https://github.com/storybookjs/storybook/blob/e0c0a1429dc123219dfd74c8c4a1c6bfc9660f74/lib/core-client/src/preview/executeLoadable.ts#L12, and this is how we are currently setting it up: https://github.com/eirslett/storybook-builder-vite/blob/fb289c1cb57c3b929eb5f586e87b7644f97e28be/packages/storybook-builder-vite/codegen-iframe-script.ts#L103-L106.

IanVS:
Fixes #201 

This enables the omission of `title` properties in the default export of stories (the new storybook "autotitle" feature).  It's a little less powerful without https://github.com/eirslett/storybook-builder-vite/pull/182, but it's still nice to support this, I think.

I did notice an issue with HMR, but it seems not to be introduced in this PR, but exists in the react example on main as well.  Changing a css value _does_ still HMR, so I think there is no regression here.

### DIFF
--- a/packages/example-react/.storybook/main.js
+++ b/packages/example-react/.storybook/main.js
@@ -6,7 +6,7 @@ module.exports = {
     builder: 'storybook-builder-vite',
   },
   features: {
-    storyStoreV7: true,
+    storyStoreV7: false,
   },
   async viteFinal(config, { configType }) {
     // customize the Vite config here
--- a/packages/example-react/package.json
+++ b/packages/example-react/package.json
@@ -15,10 +15,10 @@
     "react-dom": "17.0.2"
   },
   "devDependencies": {
-    "@storybook/addon-a11y": "^6.4.0",
-    "@storybook/addon-docs": "^6.4.0",
-    "@storybook/addon-essentials": "^6.4.0",
-    "@storybook/react": "^6.4.0",
+    "@storybook/addon-a11y": "^6.4.14",
+    "@storybook/addon-docs": "^6.4.14",
+    "@storybook/addon-essentials": "^6.4.14",
+    "@storybook/react": "^6.4.14",
     "storybook-builder-vite": "workspace:*",
     "vite": "2.7.0"
   }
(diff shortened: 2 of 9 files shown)


joshwooding:
I was seeing the issue with identical stories in the sidebar too but couldn't reproduce it :/ Hopefully #182 fixes it.

IanVS:
OK, I've rebased this onto https://github.com/eirslett/storybook-builder-vite/pull/182 (which turns out probably wasn't strictly necessary), and made a small change to what is stored as `window.STORIES`, and there are no more duplicates.  

The last issue I'm seeing is that HMR does not seem to be working correctly, and the page refreshes instead of hot reloading.

IanVS:
Hmmmm, it seems like HMR is broken and the whole page refreshes on `main` as well, when running the react example and turning off `storyStoreV7`, so I wonder if something else is going on here.

joshwooding (approved):
LGTM :)

PR #223, opened 2022-01-31
Default Cache Directory Conflict Resolution

GuyGooL5:
There is a problem starting Vite app and Storybook under storybook-builder-vite because they share the same cache directory 
 `node_modules/.vite/`  unless explicitly reassigning. 
 This PR retargets storybook-builder-vite's cache directory into a different one to not affect the app it's running along with.

### DIFF
--- a/packages/storybook-builder-vite/build.ts
+++ b/packages/storybook-builder-vite/build.ts
@@ -12,6 +12,7 @@ export async function build(options: ExtendedOptions) {
   const config = {
     configFile: false,
     root: path.resolve(options.configDir, '..'),
+    cacheDir: ".vite-storybook",
     build: {
       outDir: options.outputDir,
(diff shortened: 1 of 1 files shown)


GuyGooL5:
I guess that's a better approach yet Vite itself has it's own `.vite` folder so I thing placing the storybook cache close to it will be convenient maybe... Maybe we can even call it `.vite.storybook` and start a new convention 😅

joshwooding (approved):
LGTM

bodograumann:
Shouldn’t this also be changed in `vite-server.ts` for the `serve-storybook` command. For now it looks like it only affects the `build-storybook` mode.

IanVS:
Good catch, yes I think that would be good.  Would you like to open a pr?

PR #236, opened 2022-02-11
[fix] transitive peer deps (closes #235)

### LINKED ISSUE #235: Peer dependencies not exhaustive
Using the berry version of yarn, with the improved install logging that is default on the berry version, sb-builder-vite is being flagged as not providing react (peer dependency for @storybook/source-loader), react-dom (peer dependency for @storybook/source-loader), and typescript (peer dependency for @joshwooding/vite-plugin-react-docgen-typescript).

image

Yarn even makes the error codes clickable to open the [yarn docs page for the error code](https://yarnpkg.com/advanced/error-codes#yn0002---missing_peer_dependency).

TL;DR yarn enforces peer deps at every level of the dep tree, which results in peer dependency hoisting problems (as described by [this blog post](https://dev.to/arcanis/implicit-transitive-peer-dependencies-ed0) the yarn page linked

JamesYeoman:


IanVS:
The problem is that if we add these, then npm installs will fail for those using vue or svelte, or not using typescript.  I think we need to find another solution to this, perhaps by breaking this package up into separate plugins, for each framework or something.

JamesYeoman:
This is true. I just assumed that since @storybook/source-loader and the vite-plugin-react-docgen-typescript plugin were specified as dependencies, that this would be all that needed changing.

I've got some ideas then:
1. A new property on the viteFinal function return type to specify options like builder plugins?
2. A core builder that can delegate to plugins
3. Change @storybook/source-loader and vite-plugin-react-docgen-typescript to be optional peer dependencies

which of those sound the most sensible?

IanVS:
I'd like to hear from @eirslett and @joshwooding.  I think this is definitely something we should tackle, since we've already seen other problems related to having different storybook frameworks in this monorepo, and having a more clean separation between different frameworks seems like it might be a good solution.

JamesYeoman:
Is there any update on this? I for one feel like a plugin system could be useful going forwards, as framework-specific functionality could be handled in plugins, ultimately making the core more stable across frameworkd

joshwooding:
> I'd like to hear from @eirslett and @joshwooding. I think this is definitely something we should tackle, since we've already seen other problems related to having different storybook frameworks in this monorepo, and having a more clean separation between different frameworks seems like it might be a good solution.

I think I like the idea of a plugin system for each of the different frameworks.

JamesYeoman:
Adding onto the original reason for this PR, I've cloned the repo and ran `yarn install`, and a bunch of the transitive peer dependencies aren't fulfilled
image

For the examples at the very least, I'd say it's probably worth fixing those warnings. I'll create a separate PR to address them

JamesYeoman:
Actually, scratch that. Since I then have to pull in react, I don't know how that'll affect the vue example. I wish storybook wasn't react-specific, so that tooling could just be better across the board

IanVS:
There's a project ongoing within storybook to refactor how frameworks and builders work together.  Let's wait and see if that improves the situation a bit.  Thanks for taking the time to push a PR though, @JamesYeoman!

PR #239, opened 2022-02-11
Import story loading module from vite-app virtual module

IanVS:
I'm working on enabling Chromatic [Turbosnap](https://www.chromatic.com/docs/turbosnap) capability for vite storybook projects, and it requires a small change to the way we load in our stories to match up more closely to how the webpack builders work.  They have a separate entrypoint for just the stories, so I've moved what used to be done in the virtual `storybook-stories.js` imported directly off of the iframe, into a virtual file that's imported from the virtual `vite-app` module.  This actually more closely aligns with the way that the "modern" (storyStoreV7) code works too, which is a nice side-benefit.  

I'm going to use `patch-package` to run this code in my own storybook at work and to ensure that turbosnap works the way I am intending.  But, I wanted to put up this PR for visibility and to get any thoughts that you might have on it, @joshwooding and @eirslett.  After I prove it out for a while, I'll mark it as ready for review.

### DIFF
--- a/packages/storybook-builder-vite/code-generator-plugin.ts
+++ b/packages/storybook-builder-vite/code-generator-plugin.ts
@@ -3,7 +3,7 @@ import * as path from 'path';
 import { transformIframeHtml } from './transform-iframe-html';
 import { generateIframeScriptCode } from './codegen-iframe-script';
 import { generateModernIframeScriptCode } from './codegen-modern-iframe-script';
-import { generateImportFnScriptCode } from './codegen-importfn-script';
+import { generateImportFnScriptCode, generateVirtualStoryEntryCode } from './codegen-importfn-script';
 
 import type { Plugin } from 'vite';
 import type { ExtendedOptions } from './types';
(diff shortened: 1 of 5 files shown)


IanVS:
One other piece I haven't pushed up to this PR that's necessary for this to work, is a rollup plugin which analyzes the built files and creates a dependency graph, which chromatic can then use to determine which snapshots to re-run.  It's similar to the output stats.json that webpack can generate, but with only the parts that chromatic actually needs.  I'm uncertain whether it belongs in this repo, or (more likely) it should be a separate repo which folks can add if they want turbosnap support.  Let me know if you think it belongs here, otherwise I'll make a new project for it.

IanVS:
Here's an example of using turbosnap, by the way. 

image

It still re-runs everything in cases where it makes sense (changes to .storybook/preview.js, package.json, etc), but definitely speeds things up otherwise.

joshwooding (approved):
LGTM

IanVS:
Thanks for the review!  I'll publish the required rollup plugin next, and submit a PR to this repo to add some instructions to the README about how to use it.  Since it's not useful for anyone not using chromatic, I think it probably doesn't make sense to make a direct dependency of the builder.

PR #240, opened 2022-02-12
Fix: svelte

### LINKED ISSUE #224: Problem after updating from `0.1.13` to `0.1.14`. Sveltekit + Typescript + Storybook
Hi! After updating `storybook-builder-vite` version from `0.1.13` to `0.1.14` all stories disappear. 

console:
```Unexpected default export without title in './src/stories/Button.stories.svelte': undefined```

screenshot:
image

Example: https://github.com/SergeiZheleznov/sb-svelte-vite-example

The only thing changed is `storybook-builder-vite` version

### PR DESCRIPTION
**Pull request by bfanger**
1. Re-enabled the csfPlugin for svelte
2. Updated svelte dependencies, that fixed a `TypeError: RegisterContext is not a constructor` error. (Svelte compiler was in 'ssr' mode)

Fixes #224

(Also implemented support for preprocessors inside *.stories.svelte files)

### DIFF
--- a/packages/storybook-builder-vite/svelte/csf-plugin.ts
+++ b/packages/storybook-builder-vite/svelte/csf-plugin.ts
@@ -2,14 +2,21 @@ import { getNameFromFilename } from '@storybook/addon-svelte-csf/dist/cjs/parser
 import { readFileSync } from 'fs';
 import { extractStories } from '@storybook/addon-svelte-csf/dist/cjs/parser/extract-stories';
 const parser = require.resolve('@storybook/addon-svelte-csf/dist/esm/parser/collect-stories').replace(/[/\\]/g, '/');
+import type { Options } from "@sveltejs/vite-plugin-svelte"
+import * as svelte from "svelte/compiler"
 
-export default {
+export default function csfPlugin(svelteOptions?: Options) {
+  return {
   name: 'storybook-addon-svelte-csf',
   enforce: 'post',
-  transform(code: string, id: string) {
+  async transform(code: string, id: string) {
     if (/\.stories\.svelte$/.test(id)) {
+    
       const component = getNameFromFilename(id);
-      const source = readFileSync(id).toString();
+      let source = readFileSync(id).toString();
+      if (svelteOptions && svelteOptions.preprocess) {
+        source = (await svelte.preprocess(source, svelteOptions.preprocess)).code
+      }
       const all = extractStories(source);
       const { stories } = all;
       const storyDef = Object.entries<any>(stories)
@@ -36,4 +43,5 @@ export default {
       };
     }
   },
+}
 };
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by IanVS** [comment]
Thanks for this PR, @bfanger!  It looks like our check might need to be updated too?  

>Error: storybook-builder-vite requires @storybook/addon-svelte-csf to be installed when using @storybook/svelte.  Please install it and start storybook again.

**Comment by bfanger** [comment]
Ah, I missed adding a file in last commit. the locally compiled version of storybook-builder-vite already contained the change.

Not sure why CI fails on `@storybook/addon-svelte-csf`  running the steps from buildExamples.yml  locally (macOS) ran without error.

File: packages/storybook-builder-vite/svelte/csf-plugin.ts [hunk 1]
```diff
@@ -36,4 +42,12 @@ export default {
...
   },
+}
 };
+
+// dedent`${codeWithoutDefaultExport}
+//     const { default: parser } = require('${parser}');
+//     const __storiesMetaData = parser(${componentName}, ${JSON.stringify(storiesDef)});
+//     export default __storiesMetaData.meta;
+//     ${storyDef};
+//   `);
```
IanVS, line 53: Just curious, why is this commented out?

IanVS (changes requested):
Thanks for working on this!  When running `packages/example-svelte`, I'm seeing this error in the browser console.

```
ClientApi.js:85 Uncaught Error: Singleton client API not yet initialized, cannot call `addArgTypesEnhancer`.
```

Any idea why that might be?  

Also, I needed to add `@storybook/svelte` to the list in `optimizeDeps.ts`.  Would you mind doing that here as well?

packages/storybook-builder-vite/svelte/csf-plugin.ts
bfanger: I've removed  the comment, (I copied it from @storybook/addon-svelte-csf for reference, but in contrary to the error message, the export was not the issue.)

bfanger:
Uncaught Error: Singleton client API not yet initialized, cannot call `addArgTypesEnhancer`.  
That happens when there are multiple  versions of the @storybook/client-api installed, each with its own singleton.
#50 

After a `yarn install` I run
rm -rf ./node_modules/@storybook/svelte/node_modules/@storybook/client-api
rm -rf ./node_modules/@storybook/vue3/node_modules/@storybook/client-api

to fix that issue.

I think that's a bug in yarn, it should dedupe that dependency.

IanVS (approved):
I was really hoping somehow this would fix that issue, but yeah, I think it's some kind of yarn bug.  

But yeah, after jumping through the hoops of removing the extra deps, this works great.  Thanks for the help!

PR #244, opened 2022-02-16
fix: svelte story regex

### LINKED ISSUE #243: Incorrectly transformed some files as svelte stories
The [regex](https://github.com/eirslett/storybook-builder-vite/blob/917d8868943ec5f58c9c2c6900e196637f0d05e3/packages/storybook-builder-vite/svelte/csf-plugin.ts#L10) used to test for svelte stories is `/.stories.svelte/`, which incorrectly matches some files since the `.` means any character and there is no `$` to enforce as the file extension.

The correct regex should be `/\.stories\.[jt]sx?$/`

lokshunhung:
Resolves #243 by updating the regex to match svelte stories

### DIFF
--- a/packages/storybook-builder-vite/svelte/csf-plugin.ts
+++ b/packages/storybook-builder-vite/svelte/csf-plugin.ts
@@ -7,7 +7,7 @@ export default {
   name: 'storybook-addon-svelte-csf',
   enforce: 'post',
   transform(code: string, id: string) {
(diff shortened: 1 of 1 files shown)


IanVS (approved):
Good catch, thanks!  I think we still have other issues with Svelte which are being addressed in another PR, but this is definitely a good change to make.

PR #250, opened 2022-02-19
Fix error when no preview file is present.

joshwooding:
Currently the react example is broken due to this.


IanVS (approved):
Thanks for cleaning up after me.

PR #251, opened 2022-02-19
Move examples out of packages

joshwooding:
Pre-requisite to #257 

Currently examples are under packages. This could cause confusion when we split up the package by framework. I also find it clearer that examples are under an examples folder which mirrors the Storybook repo.


IanVS:
I haven't spent time reviewing this, but I think it's on the right track.  I'll give a more thorough look soon, but would you mind writing a description of the goal of this re-org (I believe it is so that we can support a plugin architecture for frameworks)?  I also don't know about you, but I don't have much experience with release management of a mono-repo.  Maybe some of the other storybook maintainers can give us some pointers on how we should set up our release management ([yarn release workflow](https://yarnpkg.com/features/release-workflow), lerna, something else?).

.github/workflows/buildExamples.yml
```diff
@@ -39,3 +39,28 @@ jobs:
...
+      - name: Use Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16.x
+
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
+
+      - uses: actions/cache@v2
```
IanVS, line 55: Is there an advantage to this vs. using `cache: yarn` on `setup-node` (https://github.com/actions/setup-node#caching-packages-dependencies)?

.github/workflows/buildExamples.yml
```diff
@@ -39,3 +39,28 @@ jobs:
...
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+      - name: Install dependencies
+        run: yarn install
+
+      - name: Lint
+        run: yarn lint-ci
```
IanVS, line 66: Thanks for adding this.

IanVS (approved):

.github/workflows/buildExamples.yml
joshwooding: Nope, I had a version where I swapped this for cache: yarn but I reverted it since the change wasn’t strictly necessary. Happy to change back.

PR #259, opened 2022-02-24
fix: svelte-preprocess typescript support

### LINKED ISSUE #258: *.stories.svelte does not precompile TypeScript
I have previously asked the same questions on StackOverflow:
https://stackoverflow.com/questions/70681325/storybook-vite-svelte-typescript-typescript-not-being-processed-in-st/70722451?noredirect=1#comment125676191_70722451

Where I describe how using `import type` in a `*.stories.svelte` file throws an compile error.

In the response to my question @bfanger kindly pointed out that he had made a PR that should fix my problem: #240.
However after upgrading to storybook-builder-vite@0.1.17 (latest version at the time of writing, and should include his PR), the problem still persists. (and it does not seem that this issue is the same as the one the other PR should fix: #224)

Which is why I post this issue here.

The code references has been mentioned in the StackOverflow question and have generally not changed, except I have also tried to just let my `vite.config.js` handle the preprocessing (so removing my attempt to override the preprocessing in the `.storybook/main.cjs`), but to no avail.

So my current config file files are as follows:

**.storybook/main.cjs**
```js
const { resolve } = require('path');

module.exports = {
  stories: [resolve('src/**/*.stories.@(ts|svelte)')],
  addons: ['@storybook/addon-essentials', '@storybook/addon-svelte-csf'],
  core: {
    builder: 'storybook-builder-vite',
  },
  async viteFinal(config) {
    const { default: viteConfig } = await import('../vite.config.js');

    // customize the Vite config here
    config.resolve.alias = viteConfig.resolve?.alias;

    // return the customized config
    return config;
  },
};
```

**vite.config.js**
```js
import { svelte } from '@sveltejs/vite-plugin-svelte';
import { resolve } from 'path';
import sveltePreprocess from 'svelte-preprocess';

// https://vitejs.dev/config/
/**
 * @type {import('vite').UserConfig}
 */
export default {
  plugins: [
    svelte({
      preprocess: [sveltePreprocess()],
    }),
  ],
  resolve: {
    alias: {
      '~': resolve('src'),
    },
  },
  build: {
    outDir: resolve('frontend'),
  },
};
```

Any ideas? Or is this an actual issue?

bfanger:
`if (filename == null) return { code: content };`
in [svelte-preprocess/src/transformers/typescript.ts](https://github.com/sveltejs/svelte-preprocess/blob/main/src/transformers/typescript.ts) line 497 was the issue
When you call svelte.preprocess without the "optional" filename the typescript preprocessor does nothing.

Fixes #258

@Tokimon could you verify? 
You can test it by changing your local `node_modules/storybook-builder-vite/dist/svelte/csf-plugin.js`:
Adding the `, { filename: id }` to `svelte.preprocess` call.

### DIFF
--- a/examples/svelte/.storybook/main.js
+++ b/examples/svelte/.storybook/main.js
@@ -1,3 +1,5 @@
+const preprocess = require('svelte-preprocess');
+
 module.exports = {
   framework: '@storybook/svelte',
   stories: ['../stories/**/*.stories.mdx', '../stories/**/*.stories.@(js|jsx|ts|tsx|svelte)'],
@@ -8,5 +10,8 @@ module.exports = {
   async viteFinal(config, { configType }) {
     // customize the Vite config here
     return config;
+  },
+  svelteOptions: {
+    preprocess: preprocess(),
   },
 };
--- a/examples/svelte/package.json
+++ b/examples/svelte/package.json
@@ -20,7 +20,10 @@
     "@storybook/addon-svelte-csf": "^1.1.0",
     "@storybook/svelte": "^6.4.14",
     "@sveltejs/vite-plugin-svelte": "^1.0.0-next.37",
+    "@tsconfig/svelte": "^3.0.0",
     "storybook-builder-vite": "workspace:*",
+    "svelte-preprocess": "^4.10.4",
+    "typescript": "^4.5.5",
     "vite": "2.7.0"
   }
 }
--- a/examples/svelte/stories/Button.svelte
+++ b/examples/svelte/stories/Button.svelte
@@ -1,6 +1,8 @@
-<script>
+<script lang="ts">
   import './button.css';
   import { createEventDispatcher } from 'svelte';
+  type Size = "small" | "medium" | "large" 
+  
   /**
    * Is this the principal call to action on the page?
    */
@@ -13,7 +15,7 @@
   /**
    * How large should the button be?
    */
-  export let size = 'medium';
+  export let size: Size = 'medium';
   /**
    * Button contents
    */
--- a/examples/svelte/tsconfig.json
+++ b/examples/svelte/tsconfig.json
@@ -1,0 +1,6 @@
+{
+  "extends": "@tsconfig/svelte/tsconfig.json",
+
+  "include": ["stories/**/*"],
+  "exclude": ["node_modules/*"]
+}
--- a/packages/storybook-builder-vite/svelte/csf-plugin.ts
+++ b/packages/storybook-builder-vite/svelte/csf-plugin.ts
@@ -14,7 +14,7 @@ export default function csfPlugin(svelteOptions?: Options) {
         const component = getNameFromFilename(id);
         let source = readFileSync(id).toString();
         if (svelteOptions && svelteOptions.preprocess) {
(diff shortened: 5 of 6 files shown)


IanVS:
I did a little bit of testing on this, and it seems that `svelteOptions` is always `{}`.  I'm not sure why, since we do basically the same thing as storybook itself to access it... (`await presets.apply('svelteOptions', {}, options);`).

bfanger:
That is strange, when I add a `console.log(svelteOptions)` in the csf-plugin i'm seeing:
```
svelteOptions {
  preprocess: {
    defaultLanguages: { markup: 'html', style: 'css', script: 'javascript' },
    markup: [AsyncFunction: markup],
    script: [AsyncFunction: script],
    style: [AsyncFunction: style]
  }
}
```

My storybook/main.cjs looks like:
```js
const preprocess = require("svelte-preprocess");
const { default: tsconfigPaths } = require("vite-tsconfig-paths");

module.exports = {
  core: {
    builder: "storybook-builder-vite",
  },
  stories: ["../src/**/*.stories.mdx", "../src/**/*.stories.@(ts|svelte)"],
  addons: [
    "@storybook/addon-links",
    "@storybook/addon-essentials",
    "@storybook/addon-svelte-csf",
  ],
  svelteOptions: {
    preprocess: preprocess(),
  },
  viteFinal(config) {
    /* eslint-disable no-param-reassign */
    config.base = "";
    config.plugins.push(tsconfigPaths());
    return config;
  },
};
```

IanVS:
Strange indeed.  Mind if I push up a change to our svelte example on this branch, to try to demonstrate?

bfanger:
If the change enhances the example and this PR stays mergable, yes.
Else I'd  prefer you'd push the chance into different branch (i've given you collaborate rights on my fork)

IanVS:
Thanks, I wasn't able to reproduce the problem I was having before, but did add typescript to the svelte example, so we're more likely to find problems in the future, and hopefully have an easier time debugging.  Thanks for your work here!

IanVS:
@bfanger were you able to reproduce the original problem in https://github.com/eirslett/storybook-builder-vite/issues/258?  Using the svelte example that I just modifed, I'm not able to trigger their error, even when I take out the change you made here in this PR.

bfanger:
Yes, I was able to trigger the error and verify the fix when using a simple
`const x : number = 1` inside a *.stories.svelte

Testing storybook builders is tricky:
- did I save and compile the storybook-builder-vite from ts to js
- did I delete the node_modules/.cache (just to be safe)
- did I restart the storybook dev server
- did I save the story
- did I refresh the page with the component

It's easy for miss a step, and before you know it, you're not testing what you think you're testing.

IanVS:
Agree.  Fwiw we have a watcher for compiling the builder, if you use `yarn start` at the root.   @joshwooding could you take a look at this and confirm you can trigger the error without the fix?  Otherwise, I think this looks good.

joshwooding (approved):
@IanVS I can reproduce the error and can confirm this change fixes it. Thanks @bfanger :)

Tokimon:
Thank you all for swift handling of this problem. I will make sure to test out the update :)

Tokimon:
@bfanger I did as you proposed and I got it to work as well. Thank you for treating this problem.

PR #262, opened 2022-02-25
Allow sub-dependencies in base optimizeDeps.include

### LINKED ISSUE #232: Clean install is weirdly borked
So I used `npx sb@next repro`, and it all works fine. But as I've commented in the following code block, a clean vite + sb-builder-vite install errors out

```bash
yarn create vite # I used the react-ts template
cd <folder name>
yarn
npx sb@next init --builder storybook-builder-vite # https://storybook.js.org/blog/storybook-for-vite/
yarn storybook # acorn jsx error
yarn add --dev @storybook/core-common webpack@^5 # Peer dependencies that yarn said were unmet, so this is an attempt to try to fix the error
yarn storybook # acorn jsx error still occuring
```

Here's a screenshot of the "acorn jsx error"
image

This error seems to make everything load infinitely. No error gets displayed in my shell, and everything comes to a standstill.

IanVS:
Fixes #232 

This tweaks the way we do our `optimizeDeps.include` slightly, allowing us to add `'@storybook/addon-docs > acorn-jsx'`.  I needed to change our `filter` function to use the resolution that vite does (see https://github.com/vitejs/vite/blob/67d164392e8e9081dc3f0338c4b4b8eea6c5f7da/packages/vite/src/node/optimizer/index.ts#L182-L199), but in the end I think this is more reliable than what we were doing before, anyway.  

With this syntax in place, we might also be able to avoid optimizing the storybook deps themselves, like `@storybook/addon-docs`, which already has an ESM export and doesn't really need to be pre-bundled at all, only its sub-dependencies do.  I haven't changed anything like that in this PR, though, since I wanted to take it one-step-at-a-time.  The only change made to the list itself is the addition of `'@storybook/addon-docs > acorn-jsx'`.

I also needed to prevent the vue alias from being added unless vue is actually being used, or else vite threw an error when I tried to resolve the config.

### DIFF
--- a/packages/storybook-builder-vite/optimizeDeps.ts
+++ b/packages/storybook-builder-vite/optimizeDeps.ts
@@ -1,109 +1,121 @@
 import * as path from 'path';
-import { normalizePath } from 'vite';
+import { normalizePath, resolveConfig, UserConfig } from 'vite';
 import { listStories } from './list-stories';
 
 import type { ExtendedOptions } from './types';
 
-export async function getOptimizeDeps(root: string, options: ExtendedOptions) {
+const INCLUDE_CANDIDATES = [
+  '@base2/pretty-print-object',
+  '@emotion/core',
+  '@emotion/is-prop-valid',
+  '@emotion/styled',
+  '@mdx-js/react',
+  '@storybook/addons',
+  '@storybook/addon-docs',
+  '@storybook/addon-docs > acorn-jsx',
+  '@storybook/channel-postmessage',
+  '@storybook/channel-websocket',
+  '@storybook/client-api',
+  '@storybook/client-logger',
+  '@storybook/core/client',
+  '@storybook/csf',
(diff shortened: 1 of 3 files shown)


joshwooding (approved):
Nice, definitely a better way of resolving that what was there before!

IanVS:
Hmmm, the vue example might be broken, investigating.

IanVS:
I tested all of the examples, and they're working now.  👍

eirslett:
@IanVS One reason for preoptimizing ESM dependencies is that the number of files served to the browser is lower, since it's one chunk per dependency. If we stop pre-optimizing, that could slow down Storybook performance in the browser.

IanVS:
Is that still true even for node_modules?  I thought those were still bundled into `vendor.js`.

eirslett:
I guess it's worth doing an experiment, to see what happens in the browser.

PR #263, opened 2022-02-25
Fix Node 12 support

### LINKED ISSUE #261: 0.1.17 dropped support for node 12
Our build artefacts now contains optional chaining which is only supported in Node 14+.

See https://unpkg.com/storybook-builder-vite@0.1.17/dist/code-generator-plugin.js

```js
            }
        },
        async load(id) {
            const storyStoreV7 = options.features?.storyStoreV7;
            if (id === virtualStoriesFile) {
                if (storyStoreV7) {
                    return (0, codegen_importfn_script_1.generateImportFnScriptCode)(options);
                }
                else {
                    return (0, codegen_entries_1.
```

joshwooding:
Fixes #261

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
+    "lib": ["es2019", "es2020.promise", "es2020.bigint", "es2020.string"],
     "module": "CommonJS",
-    "target": "ES2021",
+    "target": "ES2019",
     "declaration": true,
     "sourceMap": true,
     "inlineSources": true,


IanVS:
Should we include the `lib` line from here: https://github.com/tsconfig/bases/blob/0b8b889fcdada73bf3e6900bf6f74fc9d8d21b43/bases/node12.json?rgh-link-date=2022-02-25T14%3A12%3A58Z#L6?

joshwooding:
> Should we include the `lib` line from here: https://github.com/tsconfig/bases/blob/0b8b889fcdada73bf3e6900bf6f74fc9d8d21b43/bases/node12.json?rgh-link-date=2022-02-25T14%3A12%3A58Z#L6?

I don't think there's any harm.

IanVS (approved):

PR #268, opened 2022-03-01
Fix style modules not being resolved from iframe.html (#266)

### LINKED ISSUE #266: iframe style tags are stripped by Vite 2.8.5
When Vite 2.8.5 executes `transformIndexHtml` and internally runs "devHtmlHook", all the `<style>` are stripped from the iframe `<head>` resulting in Storybook's template code being exposed:

image

Have confirmed that this issue is **new for 2.8.5** (downgrading to 2.8.4 doesn't exhibit this behaviour).

Dschungelabenteuer:
Fixes #266

See comment https://github.com/eirslett/storybook-builder-vite/issues/266#issuecomment-1055677865

* Upgraded to Vite 2.8.5 to reproduce the issue
* Updated outdated CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,7 @@ In another Terminal run:
 yarn example:NAME
 ```
 
-Where `NAME` is based on a `packages/example-*` directory.
+Where `NAME` is based on a `examples/*` directories.
 
 For Vue:
 
--- a/examples/react-ts/package.json
+++ b/examples/react-ts/package.json
@@ -21,6 +21,6 @@
     "@storybook/react": "6.4.14",
     "storybook-builder-vite": "workspace:*",
     "typescript": "^4.5.4",
-    "vite": "2.7.0"
+    "vite": "2.8.5"
   }
 }
--- a/examples/react/package.json
+++ b/examples/react/package.json
@@ -20,6 +20,6 @@
     "@storybook/addon-essentials": "^6.4.14",
     "@storybook/react": "^6.4.14",
     "storybook-builder-vite": "workspace:*",
-    "vite": "2.7.0"
+    "vite": "2.8.5"
   }
 }
--- a/examples/svelte/package.json
+++ b/examples/svelte/package.json
@@ -24,6 +24,6 @@
     "storybook-builder-vite": "workspace:*",
     "svelte-preprocess": "^4.10.4",
     "typescript": "^4.5.5",
-    "vite": "2.7.0"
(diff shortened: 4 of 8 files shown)


IanVS:
Thanks for digging into this!  It sounds like vite is going to revert for now, and re-introduce this change into 2.9 (https://github.com/vitejs/vite/pull/7136).  So, let's hold off on this change for now, until things stabilize a bit.  But this seems like a reasonable fix once they do make the change.

Dschungelabenteuer:
@IanVS I've seen the discussion on Vite's Discord server, should I close this PR or do you want me to update it to include poyoho's implementation of the fix ?

IanVS:
If you're up for using that approach, that would be great!  Although, it would be nice if we can use native browser methods and avoid adding a dependency on 'url'.  Thanks!

Dschungelabenteuer:
Hey! Just gave it a try and it seems to be working fine! However, using the native URL API requires:
* adding `"DOM"` and `"DOM.Iterable"` to the _tsconfig_'s `lib`   property
* a valid URL (which `/iframe.html` is not), so my first idea was to mock a local file, but I'm not sure this is the best way to go

Any suggestions? [See changes](https://github.com/Dschungelabenteuer/storybook-builder-vite/pull/1/files)

IanVS:
Thanks yet again!  I think I need to apologize for sending you down the wrong track.  I'm so used to working in the browser that sometimes I forget that most of this code runs in node.  And I didn't realize that `url` is a native node module, so `import { parse as parseUrl, URLSearchParams } from 'url';`  isn't actually adding a dependency, but rather using the node module.  If you used that approach, would it simplify things?  Again, sorry for sending you down the wrong track there.

Dschungelabenteuer:
Hey Ian! No worries, I should have been able to spot that myself 😄

Node URL module's `parseUrl` is however marked as deprecated since it's part of the [legacy API](https://nodejs.org/dist/latest-v8.x/docs/api/url.html#url_legacy_url_api). Using the WHATWG URL API - which mimics the browser URL API - still requires me to add a fake protocol for the URL to be valid. What would be your take on this ?

Dschungelabenteuer:
Sounds fine to me, here you go 🤞

IanVS:
It looks like another `yarn install` is needed, there seem to be changes to yarn.lock.

IanVS (approved):
Thanks!

IanVS:
Released in 0.1.19.

PR #269, opened 2022-03-03
Add `jest-mock` to optimized deps

### LINKED ISSUE #242: Export name issue with jest-mock while running Storybook
Hello there,

When I install Vite then `storybook-builder-vite` with `npx sb@latest init --builder storybook-builder-vite`, the compilation works fine but then Storybook fails to start throwing this error:
Screenshot 2022-02-16 at 11 33 44

I have tried to add `jest-mock` to the `optimizeDeps`, install the dependency but nothing works.

When I run `npx sb@next init --builder storybook-builder-vite`, I have a similar error with `acorn-jxs` which has a missing export as well.
I suspect that the issue is related to ESM and how the dependencies exports their methods.
Any clue or help to solve this issue?

Thanks!

IanVS:
Fixes #242.

This is getting pulled in by `@storybook/addon-interactons`, which is scaffolded into new storybook projects.

### DIFF
--- a/packages/storybook-builder-vite/optimizeDeps.ts
+++ b/packages/storybook-builder-vite/optimizeDeps.ts
@@ -39,6 +39,7 @@ const INCLUDE_CANDIDATES = [
   'global',
   'html-tags',
   'isobject',
+  'jest-mock',
   'loader-utils',
   'lodash/cloneDeep',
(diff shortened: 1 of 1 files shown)


joshwooding (approved):
LGTM

PR #274, opened 2022-03-14
Test built storybooks in CI

IanVS:
So far, we've only been building our storybooks, but not doing any testing on them.  This adds https://github.com/storybookjs/test-runner which will render each of the stories and smoke test them (make sure that there are no runtime errors).  This is at least a step towards having more confidence that our stories aren't broken.  The next step would be to actually write assertions in the stories to be sure they're rendering _correctly_, but that can be done another time.

### DIFF
--- a/.github/workflows/buildExamples.yml
+++ b/.github/workflows/buildExamples.yml
@@ -27,7 +27,7 @@ jobs:
         run: yarn prepublish
 
       - name: Build examples
-        run: yarn workspaces foreach -p run build-storybook
+        run: yarn build-examples
   lint:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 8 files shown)


IanVS:
The svelte stories are not being picked up correctly by the test runner.  I've opened https://github.com/storybookjs/test-runner/issues/80.

IanVS:
Hmmm, there seem to be other issues in vue and svelte for the alpha versions, which I'll need to investigate later.

IanVS:
I had to remove the svelte example, unfortunately.  It seems that @storybook/test-runner does not support svelte component stories.

eirslett:
LGTM!

PR #276, opened 2022-03-14
Use `req.query` instead of custom parsing.

joshwooding:


packages/storybook-builder-vite/index.ts
```diff
@@ -16,25 +15,16 @@ export interface ViteStats {}
...
       next();
       return;
     }
 
-    const query = parseRequest(req.url)!;
-
-    if (query['html-proxy'] !== undefined) {
+    // We need to handle `html-proxy` params for style tag HMR https://github.com/eirslett/storybook-builder-vite/issues/266#issuecomment-1055677865
+    // e.g. /iframe.html?html-proxy&index=0.css
+    if (req.query['html-proxy'] !== undefined) {
```
joshwooding, line 27: I kept this separate to the check above for readability. Happy to merge them and move the comment if it's preferred.

IanVS (approved):
If I had known about `req.query`, I could have saved poor @Dschungelabenteuer a lot of work.  😭

packages/storybook-builder-vite/index.ts
IanVS: It's good this way, I think.

PR #277, opened 2022-03-14
Fix handling of rejected promises when attempting to close the server

Codex-:
Current:
```ts
function bail(e?: Error) {
  try {
    returnserver.close();
  } catch (err) {
    console.warn('unable to close vite server');
  }

  throw e;
}
```

If this call to `server.close()` throws, then it does not get caught by the `catch` here as the promise is not awaited. For this catch to handle the rejection of this promise, the `server.close()` call needs to be `await`ed


IanVS:
Thanks!

PR #284, opened 2022-03-19
Bump @joshwooding/vite-plugin-react-docgen-typescript to v0.0.4

joshwooding:


IanVS (approved):
It looks like https://github.com/joshwooding/vite-plugin-react-docgen-typescript/commit/85247cfeb3f7bd59c67f564053c1f0b18247bc49 is the change here.  This _does_ remove the warningsof  "Sourcemap is likely to be incorrect: a plugin (vite:react-docgen-typescript) was used to transform files, but didn't generate a sourcemap for the transformation. Consult the plugin documentation for help", but doesn't actually generate any source maps.  I think this is correct for this logic branch, since it's returning the original src.  But, it also returns `map: null` when it _does_ transform the source.  It looks like that's what it's always done, and we haven't heard any complaints, but I do wonder if we should be generating a sourcemap in that situation.

IanVS (approved):
I've convinced myself this is fine, since the new code is all added at the end.

IanVS (approved):

IanVS:
Nice, the new version includes sourcemap support!  https://discord.com/channels/@me/913167615625687080/959867321856434236

PR #297, opened 2022-03-31
Add migration to readme, other updates

IanVS:
This makes a few tweaks to our project readme:

1) Adds a `## Project has been renamed` section, with instructions how to migrate.
2) Moves the call for maintainers to the ## Contributing section, to avoid making it seem like this is a struggling project.
3) Adds pnpm installation instructions, including shameful hoisting, and moves Requirements into the installation section.
4) Updates the "for a new project" section to use `npm create vite@latest`.
5) Removes a few mentions of known issues that have been fixed. \o/


IanVS:
Going to self-merge this one, since it's just docs cleanup. Feel free to put up another PR if you think anything should change.

PR #305, opened 2022-04-03
Append browserHash to absolute config imports

IanVS:
This reverts commit https://github.com/storybookjs/builder-vite/pull/289, and fixes https://github.com/storybookjs/storybook/issues/17773 a different way.

@tmeasday and I determined that the `start()` function in [app/vue3/src/client/preview/index.ts](https://github.com/storybookjs/storybook/blob/534e61855717fea1ff7ce554523a4d59ff453553/app/vue3/src/client/preview/index.ts#L25) was being called twice, once with all the parameters, and once with just the framework renderer.  Eventually I realized that one of the urls calling start was something like `http://localhost:6006/node_modules/@storybook/vue3/dist/esm/client/preview/index.js?v=afbc4050` and the other was `http://localhost:6006/node_modules/@storybook/vue3/dist/esm/client/preview/index.js` (notice no `v=browserHash` at the end).  It seems this is due to several things:

- Using the `/@fs/` syntax to point to a file on the filesystem, rather than a node_module (maybe vite doesn't automatically append the browser hash in this case?  Maybe @patak-dev can confirm.)
- Pre-bundling several `@storybook` packages, even though there were esm exports available to import.  I'm not sure why these are being pre-bundled, but excluding them was necessary to avoid the double-start.

The "fix" here is a bit of a hack, and I'd like to find a better way.  I'm grabbing the `browserHash` from a private property of the server (not sure what versions of vite this even works with), and manually appending it to the filenames that we're generating.  That alone wasn't enough to fix the problem, though, as I also needed to prevent the `@storybook` packages from being optimized.  

Lastly, this only works if one more change is made to storybook frameworks, to change [this line](https://github.com/storybookjs/storybook/blob/534e61855717fea1ff7ce554523a4d59ff453553/app/vue3/src/client/preview/index.ts#L2) (and it's equivalent in each framework) from 

```js
import { start } from '@storybook/core/client';
```

to 

```js
import { start } from '@storybook/core-client';
```

Because `@storybook/core/client.js` is a cjs file.

I hope that maybe some of the vite folks can help suggest a better way than manually appending a browserHash.  I may need to put together a simpler reproduction of the issue that better demonstrates what's going on, but I'm not convinced I'll be able to.

Or, maybe we can change / resolve the absolute paths into normal bare imports and allow the typical import resolution to occur?  I.e. change from:

```js
import * as config_1 from '/@fs//Users/ianvs/code/experiments/sb-builder-vite-no-container/node_modules/@storybook/vue3/dist/esm/client/preview/config'
```

to something like:

```js
import * as config_1 from '@storybook/vue3/client/preview/config'
```

But I think that would require a change to core storybook, and might break the webpack builders.

### DIFF
--- a/packages/builder-vite/code-generator-plugin.ts
+++ b/packages/builder-vite/code-generator-plugin.ts
@@ -15,6 +15,7 @@ import { virtualAddonSetupFile, virtualFileId, virtualPreviewFile, virtualStorie
 export function codeGeneratorPlugin(options: ExtendedOptions): Plugin {
   const iframePath = path.resolve(__dirname, '..', 'input', 'iframe.html');
   let iframeId: string;
+  let browserHash: string;
 
   // noinspection JSUnusedGlobalSymbols
   return {
@@ -87,7 +88,7 @@ export function codeGeneratorPlugin(options: ExtendedOptions): Plugin {
         if (storyStoreV7) {
           return generateModernIframeScriptCode(options);
         } else {
-          return generateIframeScriptCode(options);
+          return generateIframeScriptCode({ ...options, browserHash });
         }
       }
 
@@ -99,6 +100,9 @@ export function codeGeneratorPlugin(options: ExtendedOptions): Plugin {
       if (ctx.path !== '/iframe.html') {
         return;
       }
+      // We need the browserHash from the server so that we can append it to imports.
+      // @ts-expect-error accessing private property of server
+      browserHash = ctx.server?._optimizedDeps.metadata.browserHash ?? '';
       return transformIframeHtml(html, options);
     },
   };
--- a/packages/builder-vite/codegen-entries.ts
+++ b/packages/builder-vite/codegen-entries.ts
@@ -9,17 +9,14 @@ const absoluteFilesToImport = (files: string[], name: string) =>
   files.map((el, i) => `import ${name ? `* as ${name}_${i} from ` : ''}'/@fs/${normalizePath(el)}'`).join('\n');
 
 export async function generateVirtualStoryEntryCode(options: ExtendedOptions) {
-  const { frameworkPath, framework } = options;
   const storyEntries = await listStories(options);
   const resolveMap = storyEntries.reduce<Record<string, string>>(
     (prev, entry) => ({ ...prev, [entry]: entry.replace(slash(process.cwd()), '.') }),
     {}
   );
   const modules = storyEntries.map((entry, i) => `${JSON.stringify(entry)}: story_${i}`).join(',');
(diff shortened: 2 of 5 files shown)


IanVS:
To be clear, this is blocked by imports of `@storybook/core/client`, and even if that change is made, I'm hoping we can find a better way.

patak-dev:
> Eventually I realized that one of the urls calling start was something like http://localhost:6006/node_modules/@storybook/vue3/dist/esm/client/preview/index.js?v=afbc4050 and the other was http://localhost:6006/node_modules/@storybook/vue3/dist/esm/client/preview/index.js (notice no v=browserHash at the end).

This may be a bug in Vite. Could you point me to a way to reproduce it?

tmeasday:
A suggestion for a repro @IanVS:

- Make two modules `a` and `b`.
- import `a` from `b`
- create a virtual entry that imports both `a` and `b` without a hash. 

What we saw was that when internal imports crossed module boundaries, the hash was appended by vite. So I would expect `b`'s import of `a` to have the hash, but the virtual entries to not. 

Then make `a` do something singleton-y :)

IanVS:
@patak-dev, here is as minimal repro as I can make.  https://github.com/IanVS/vite-fs-import-prebundle-repro

Perhaps this is just expected behavior, and I need to find a different way to resolve a file on filesystem into a node_module module that might have been pre-bundled?

IanVS:
Thanks @patak-dev.  Do you know what the docs are referring to here? https://vitejs.dev/config/#server-fs-allow.  

>Restrict files that could be served via `/@fs/`

It seems like direct access via `/@fs/` is allowed in some cases, but maybe not node modules?

@tmeasday, why are [these](https://github.com/storybookjs/storybook/blob/534e61855717fea1ff7ce554523a4d59ff453553/app/vue3/src/server/framework-preset-vue3.ts#L52) set up as absolute paths rather than perhaps module names which could be resolved as normal imports?  I'm not really sure if there's a way we can convert from `/xxx/node_modules/@storybook/vue3/dist/esm/client/preview/config` to `@storybook/vue3/client/preview/config`, is there?  Other than a hacky and brittle string-replacement, I guess, which might be what's necessary for now...

tmeasday:
I believe we need to use the full path to `dist/esm` as an entry to the package. 

But I guess you are really asking about the absolute path. @ndelangen or @shilman can probably speak to how necessary that is.

tmeasday:
To hazard a guess I would say it's due to tools like pnpm which don't let you require a package you don't depend on unless you provide a absolute path (do I have that right?)

PR #313, opened 2022-04-04
Check for explicit typescript dep

### LINKED ISSUE #312: [Bug] Detection of typescript for react-docgen is broken
### Describe the bug

We currently [check](https://github.com/storybookjs/builder-vite/blob/bf71b58ab59029afa4198394c5fa12b65a799b08/packages/builder-vite/vite-config.ts#L110-L115) to see if typescript is installed to determine whether to default to react-docgen-typescript or react-docgen.  However, typescript comes along for the ride it seems, no matter whether it's used in the project or not:

```
npm why typescript
typescript@4.6.3 dev peer
node_modules/typescript
  peer typescript@">= 2.7" from fork-ts-checker-webpack-plugin@6.5.0
  node_modules/fork-ts-checker-webpack-plugin
    fork-ts-checker-webpack-plugin@"^6.0.4" from @storybook/core-common@6.4.20
```

So, I think we need to find a different way to look for typescript.  Perhaps by checking the `package.json` for typescript, or some other way.  I wonder how the webpack builder handles this...

### Steps to reproduce the behavior

Create a non-typescript react vite project, initialize storybook, then print out `config` in viteFinal, you'll see that `vite:react-docgen-typescript` is being used.

### Expected behavior

In a non-typescript project, we should use `react-docgen` instead.

IanVS:
Fixes #312.

Previously, we'd just look to see if typescript was in the project at all, but turns out that it can be included as a peer dependency of some storybook dependencies.

So, this tightens the check a bit, to search the `package.json` for `typescript` in either `devDependencies` or `dependencies` (in case someone forgot to use `--save-dev` I guess).  

To verify, I removed the `typescript` config from our react example (that should have been a clue to me this was broken, I assumed it was due to our monorepo), and I also checked that `vite:react-docgen-typescript` is still being used in our react-ts example.

### DIFF
--- a/examples/react/.storybook/main.js
+++ b/examples/react/.storybook/main.js
@@ -12,9 +12,4 @@ module.exports = {
     // customize the Vite config here
     return config;
   },
-  // Normally this wouldn't be necessary, but `react-docgen-typescript` is the default,
-  // and since typescript is installed in the monorepo, we can't rely on that check.
-  typescript: {
-    reactDocgen: 'react-docgen',
-  },
 };
--- a/packages/builder-vite/vite-config.ts
+++ b/packages/builder-vite/vite-config.ts
@@ -1,4 +1,5 @@
 import * as path from 'path';
+import fs from 'fs';
 import { Plugin } from 'vite';
 import { allowedEnvPrefix as envPrefix } from './envs';
 import { TypescriptConfig } from '@storybook/core-common';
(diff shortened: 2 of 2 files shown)


IanVS:
I'd really like to get this out in the next release, and it seems to be working correctly, so I'm going to go ahead and self-merge, but I'd still love your 👀 on this at some point, @joshwooding, in case you can think of improvements to make.

IanVS:
That's a good point.  Can you think of any other ways to handle this?  I guess I'd rather error on the side of not detecting it, which means react-docgen is used, which still works, rather than a false positive, which means react-docgen-typescript is used on a non-typescript repo, which breaks.

PR #315, opened 2022-04-06
chore(docs): update readme to use mergeConfig vite helper function

benatshippabo:
## Summary
Recommend users to use the [`mergeConfig` vite helper function](https://github.com/vitejs/vite/blob/77dc1a19b4f6f079a3861773b63e36177f32504d/packages/vite/src/node/config.ts#L768-L774) to easily recursively merge config options without troublesome nesting and object spreading shallow copy pitfalls.


IanVS:
I love it, thanks for the suggestion!  Would you mind running `yarn lint:prettier` and pushing up the result?

PR #324, opened 2022-04-08
Correctly import/prebundle V6 store configs

IanVS:
Closes https://github.com/storybookjs/builder-vite/pull/305

This reverts https://github.com/storybookjs/builder-vite/pull/289, and fixes it a better way, by replacing filesystem imports (`/@fs/` style absolute path imports) with normal node_modules style imports.  So, instead of:

```
/@fs/Users/ianvs/code/experiments/sb-builder-vite-no-container/node_modules/@storybook/vue3/dist/esm/client/preview/config
```

We'll import from 

```
@storybook/vue3/dist/esm/client/preview/config
```

Which vite can correctly pre-bundle and add a browserHash to.

To test this, run the vue example in this repo.  You should see some messages about new dependencies being optimized in the terminal, and the storybook should load up just fine with no errors about `No docs.container set`.

I tested this out back to our minimum supported versions of vite (2.5.2) and storybook (6.4.0).  It worked fine, although at that old version, one page refresh was required on the very first startup (due to changes in pre-bundled deps).  This doesn't happen in vite 2.9.0, which has better support for finding new deps.

### DIFF
--- a/packages/builder-vite/codegen-entries.ts
+++ b/packages/builder-vite/codegen-entries.ts
@@ -9,17 +9,14 @@ const absoluteFilesToImport = (files: string[], name: string) =>
   files.map((el, i) => `import ${name ? `* as ${name}_${i} from ` : ''}'/@fs/${normalizePath(el)}'`).join('\n');
 
 export async function generateVirtualStoryEntryCode(options: ExtendedOptions) {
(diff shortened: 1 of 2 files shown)


shilman (approved):
LGTM! 💪

joshwooding (approved):
LGTM

PR #338, opened 2022-04-15
react-docgen: exclude node_modules, make configurable

IanVS:
This updates the internal react-docgen plugin to prevent running against files that are in node_modules ~or outside of the project root~.

I realized we should do this when working on converting a large project over to use the vite builder as an experiment, and found that it took ~20 minutes to load the first storybook page after the browser opened.  After digging, I found that most of this time was spent processing files that were in other parts of the monorepo, like icons, which did not need docgen info.  After making the changes in this PR, the time was reduced to less than 2 minutes on a cold start.  The number of processed files dropped from 1668 to 958.  And the time drops to 20 seconds for a cold boot with storyStoreV7 (processes 39 files).

I'm making a few assumptions here:

1) Components in node_modules should not get docgen information.  I think this is reasonable, since it's unlikely you need to document imported components, they should be documented elsewhere.

~2) Components outside of your project root (technically, I'm using `process.cwd()` here)  should likewise not get docgen.  This may not always be a good assumption to make, and by supplying this exclude pattern as a default parameter, I'm thinking we could easily pull this plugin into a separate package and allow users to customize it.  But, more often than not, I think this is the right decision to exclude them.  Curious to hear other's thoughts though.~

This is configurable, although it's not super-straightforward.  For example:

```js
async viteFinal(config) {
    config.plugins = [
      require('@storybook/builder-vite/dist/plugins/react-docgen').reactDocgen({
        exclude: [/^\.\.\//, /node_modules/],
        include: /\.(jsx?)$/,
      }),
      ...config.plugins.filter((plugin) => plugin?.name !== 'react-docgen'),
    ];
    return config;
}
```

Note that `react-docgen` needs to happen before the `@vitejs/plugin-react` plugins, which is why it's above the spread of the rest of the plugins in this example.

### DIFF
--- a/packages/builder-vite/plugins/react-docgen.ts
+++ b/packages/builder-vite/plugins/react-docgen.ts
@@ -24,9 +24,9 @@ type Options = {
   exclude?: string | RegExp | (string | RegExp)[];
 };
 
-// pattern /^..\// prevents files that are outside of the project root (process.cwd());
-const DEFAULT_EXCLUDE = [/node_modules\/.*/, /^..\//];
-export function reactDocgen(options: Options = { include: /\.(mjs|tsx?|jsx?)$/, exclude: DEFAULT_EXCLUDE }): Plugin {
+export function reactDocgen(
+  options: Options = { include: /\.(mjs|tsx?|jsx?)$/, exclude: [/node_modules\/.*/] }
+): Plugin {
   const cwd = process.cwd();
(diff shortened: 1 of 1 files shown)


IanVS:
This breaks the workspace example.  I need to find another way to handle this, perhaps we can just keep the plugin configurable and if a project needs to modify it, they can use `viteFinal` to adjust it.

joshwooding (approved):

PR #346, opened 2022-04-22
Avoid hiding all non-storybook body children

IanVS:
Fixes https://github.com/storybookjs/builder-vite/issues/343

This replaces the blunt hammer of 

```css
body > * {
  display: none !important;
}
```

with something a bit more nuanced, taken from the storybook default head styles.  

This is only necessary until https://github.com/vitejs/vite/pull/7786 is released, then we can remove the workaround and ask users to update vite in order to avoid this flash.

### DIFF
--- a/packages/builder-vite/plugins/no-fouc.ts
+++ b/packages/builder-vite/plugins/no-fouc.ts
@@ -21,22 +21,27 @@ export function noFouc(): Plugin {
   };
 }
 
+/**
+ * Insert default styles to hide storybook elements as the page loads until JS can
+ * add the official storybook default head styles and scripts.  These lines are mostly
+ * taken from https://github.com/storybookjs/storybook/blob/next/lib/core-common/templates/base-preview-head.html#L6-L20
+ */
 function insertHeadStyles(html: string) {
   return html.replace(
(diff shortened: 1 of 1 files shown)


IanVS:
@eirslett @joshwooding while exploring this, I realized that the `html` provided to `transformIndexHtml` in `code-generator-plugin.ts` is already transformed, due to `iframeMiddleware` in the `index.ts` file.  Do either of you know why it's included in `code-generator-plugin.ts`, given this is the case?

joshwooding:
> @eirslett @joshwooding while exploring this, I realized that the `html` provided to `transformIndexHtml` in `code-generator-plugin.ts` is already transformed, due to `iframeMiddleware` in the `index.ts` file. Do either of you know why it's included in `code-generator-plugin.ts`, given this is the case?

I'm not too sure 😕

joshwooding (approved):

IanVS:
Going to merge, but I'd still be curious to hear if Eirik knows the reason for this.  I think we can probably remove it in a separate PR.

eirslett:
I'm not sure 🤷 it's probably fine to change it.

Issue #329, opened 2022-04-10
[Bug] StorybookConfig type does not recognize `viteFinal` as a property


bilalq:
### Describe the bug

I've written my Storybook config in TypeScript as a `main.ts` file instead of a `main.js` file, but the typings of `StorybookConfig` do not recognize `viteFinal` as a valid property.

```
[tsserver 2322] [E] Type '{ stories: string[]; staticDirs: string[]; addons: string[]; framework: string; core: { builder: string; }; features: { storyStoreV7: true; }; viteFinal(storybookDefaultConfig: Record<string, any>, { configType: _configType }: { ...; }): Promise<...>; }' is not assignable to type 'StorybookConfig'.
  Object literal may only specify known properties, and 'viteFinal' does not exist in type 'StorybookConfig'.
```

The definition for those types actually live in [@storybook/core-common](https://github.com/storybookjs/storybook/blob/next/lib/core-common/src/types.ts#L311). I wasn't sure whether to create this issue in the main repo or here, but figured this was more applicable.

### Steps to reproduce the behavior

1. Create a clean vite project
2. Initialize latest storybook setup with `npx sb@next init --builder @storybook/builder-vite`
3. Rename `main.js` to `main.ts` and try to assign the config object to the type of `StorybookConfig`. 
4. See type error.

Here is my example config in `main.ts`:

```typescript
import type { StorybookConfig } from '@storybook/core-common'
import { mergeConfig } from 'vite'
import appViteConfig from '../vite.config'

const config: StorybookConfig = {
  stories: ['../src/**/*.stories.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx)'],
  staticDirs: ['../public'],
  addons: [
    '@storybook/addon-links',
    '@storybook/addon-essentials',
    '@storybook/addon-interactions',
  ],
  framework: '@storybook/react',
  core: {
    builder: '@storybook/builder-vite',
  },
  features: {
    storyStoreV7: true,
  },
  async viteFinal(
    storybookDefaultConfig: Record<string, any>,
    { configType: _configType }
  ) {
    // We combine Storybook's default Vite config with our own. Any Storybook
    // specific extensions are customized here as well.
    return mergeConfig(mergeConfig(storybookDefaultConfig, appViteConfig), {
      optimizeDeps: {
        // Including things here works around Vite/storybook prebundling issues
[...]

PR #349, opened 2022-04-22
Add ViteFinal types

dohooo:
Now I can use this type like this
```
const config: StorybookConfig & ViteFinal = {
    viteFinal(config, options) {

    },
}
```
#329

### DIFF
--- a/packages/builder-vite/README.md
+++ b/packages/builder-vite/README.md
@@ -113,7 +113,7 @@ import { StorybookViteConfig } from "@storybook/builder-vite"
 
 const config: StorybookViteConfig = {
     ...,
-    async viteFin(config, options) {
+    async viteFinal(config, options) {
       // do something
     },
 }
--- a/packages/builder-vite/README.md
+++ b/packages/builder-vite/README.md
@@ -113,26 +113,22 @@ import { StorybookViteConfig } from "@storybook/builder-vite"
 
 const config: StorybookViteConfig = {
     ...,
-    viteFinal(config, options) {
+    async viteFin(config, options) {
       // do something
     },
 }
+
+export default config;
 ```
 
 or
 
 ```typescript
-import { StorybookViteConfig, ViteFinal } from "@storybook/builder-vite"
+import { StorybookViteConfig, ViteFinal } from '@storybook/builder-vite';
 
-const viteFinal: ViteFinal = (config, options) => {
-    // do something
-}
-
-const config: StorybookViteConfig = {
-    ...,
-    viteFinal
-}
-
+export const viteFinal: ViteFinal = async (config, options) => {
+  // do something
+};
 ```
 
 ## Note about working directory
--- a/packages/builder-vite/README.md
+++ b/packages/builder-vite/README.md
@@ -104,6 +104,37 @@ module.exports = {
 };
 ```
 
+## TypeScript
+
+- Configure your `.storybook/main.ts` to use TypeScript
+
+```typescript
+import { StorybookViteConfig } from "@storybook/builder-vite"
+
+const config: StorybookViteConfig = {
+    ...,
+    viteFinal(config, options) {
+      // do something
+    },
+}
+```
+
+or
+
+```typescript
+import { StorybookViteConfig, ViteFinal } from "@storybook/builder-vite"
+
+const viteFinal: ViteFinal = (config, options) => {
+    // do something
+}
+
+const config: StorybookViteConfig = {
+    ...,
+    viteFinal
+}
+
+```
+
 ## Note about working directory
 
 The builder will by default enable Vite's [server.fs.strict](https://vitejs.dev/config/#server-fs-strict)
--- a/packages/builder-vite/index.ts
+++ b/packages/builder-vite/index.ts
@@ -6,14 +6,18 @@ import { transformIframeHtml } from './transform-iframe-html';
 import { createViteServer } from './vite-server';
 import { build as viteBuild } from './build';
 
-import type { Builder } from '@storybook/core-common';
+import type { Builder, Options } from '@storybook/core-common';
 import type { RequestHandler, Request, Response } from 'express';
-import type { UserConfig, ViteDevServer } from 'vite';
+import type { InlineConfig, UserConfig, ViteDevServer } from 'vite';
 import type { ExtendedOptions } from './types';
 
 export interface ViteStats {}
 
 export type ViteBuilder = Builder<UserConfig, ViteStats>;
+
+export type ViteFinal = {
+  viteFinal(config: InlineConfig, options: Options): Record<string, any>;
+};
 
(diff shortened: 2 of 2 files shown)


packages/builder-vite/index.ts
```diff
@@ -6,15 +6,19 @@ import { transformIframeHtml } from './transform-iframe-html';
...
+import type { InlineConfig, UserConfig, ViteDevServer } from 'vite';
 import type { ExtendedOptions } from './types';
 
 export interface ViteStats {}
 
 export type ViteBuilder = Builder<UserConfig, ViteStats>;
 
+export type ViteFinal = {
+  viteFinal(config: InlineConfig, options: Options): Record<string, any>;
+};
```
bilalq, line 20: This type isn't quite correct. It should be a function that returns a `InlineConfig | Promise<InlineConfig>`, not a `Record<string, any>`. Also, this is a bit of a personal nit, but it's a bit awkward for the wrapper type to be named after the function it contains. 

Something like this would be more accurate. 
```typescript
export type ViteFinal = (
  config: InlineConfig,
  options: Options
) => InlineConfig | Promise<InlineConfig>;
```

Also, I have a bit of an opinionated take here: Let's just export a type that's ready to use rather than having the user import from various places and intersect on their own.

```suggestion
export type StorybookViteConfig = StorybookConfig & {
  viteFinal: (
    config: InlineConfig,
    options: Options
  ) => InlineConfig | Promise<InlineConfig>;
};
```

packages/builder-vite/index.ts
```diff
@@ -6,15 +6,19 @@ import { transformIframeHtml } from './transform-iframe-html';
 import { createViteServer } from './vite-server';
 import { build as viteBuild } from './build';
 
-import type { Builder } from '@storybook/core-common';
+import type { Builder, Options } from '@storybook/core-common';
```
bilalq, line 9: This suggestion is intended to pair with my other suggestion comment. 

```suggestion
import type { Builder, Options, StorybookConfig } from '@storybook/core-common';
```

bilalq:
Thanks for working on this! I left some feedback for what I think would be improvements.

packages/builder-vite/index.ts
dohooo: Does some error occur when the StorybookConfig type referenced by Builder-Vite itself is inconsistent with the user version, Wouldn't it be better to export it in a single type? 
>  user import from various places and intersect on their own
It's very cheap to change

dohooo: Thank you for your advice!~

bilalq: Would that problem actually happen? The `@storybook/core-common` package is a peer dep, so in theory, only the user's version loads, right? Or am I misunderstanding something? Is it a transitive dep?

dohooo: oh my bad，thx for your remind

dohooo:
All resolved.

Asim-Tahir (approved):

IanVS:
Looks good to me as well.  Would you like to add a section to the README so that others can learn how to use this as well?  I think a dedicated `Typescript` section makes sense, since there's a bit more I can think of that I'll add in there later.

dohooo:
> Looks good to me as well. Would you like to add a section to the README so that others can learn how to use this as well? I think a dedicated `Typescript` section makes sense, since there's a bit more I can think of that I'll add in there later.

Updated~

packages/builder-vite/README.md
```diff
@@ -104,6 +104,37 @@ module.exports = {
...
+## TypeScript
+
+- Configure your `.storybook/main.ts` to use TypeScript
+
+```typescript
+import { StorybookViteConfig } from "@storybook/builder-vite"
+
+const config: StorybookViteConfig = {
+    ...,
+    viteFinal(config, options) {
```
IanVS, line 116: Let's make this async, to match with our other examples.

packages/builder-vite/README.md
```diff
@@ -104,6 +104,37 @@ module.exports = {
...
 ```
 
+## TypeScript
+
+- Configure your `.storybook/main.ts` to use TypeScript
+
+```typescript
+import { StorybookViteConfig } from "@storybook/builder-vite"
+
+const config: StorybookViteConfig = {
```
IanVS, line 114: Doesn't this need to be default exported at the end?

IanVS (commented):
I have a few small comments, but if you're tired with the back-and-forth, I can merge this and tweak it afterwards if you prefer.  :)

packages/builder-vite/README.md
dohooo: Like `async viteFinal(...` ？

IanVS: Yep, exactly.

packages/builder-vite/README.md
dohooo: Oh, I mean just doing the basic TS usage... I'll add it

IanVS: Thanks.  I think sometimes people copy-paste examples and then can get confused when they don't work directly.  :)

packages/builder-vite/README.md
```diff
@@ -104,6 +104,37 @@ module.exports = {
...
+import { StorybookViteConfig, ViteFinal } from "@storybook/builder-vite"
+
+const viteFinal: ViteFinal = (config, options) => {
+    // do something
+}
+
+const config: StorybookViteConfig = {
+    ...,
+    viteFinal
+}
```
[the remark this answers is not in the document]
dohooo: It can get types' hints if do so. I not sure what you mean.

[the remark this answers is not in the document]
IanVS: I think both versions you show here should provide type hints, right?  The advantage I see in this second style  is that you can

```ts
export const viteFinal: ViteFinal = async (config, options) => {
    // do something
}
```

And not need the `const config` at all.

But, I'm not sure how many people use this style.

Asim-Tahir (approved):

packages/builder-vite/README.md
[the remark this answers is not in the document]
dohooo: WOW！ I don't it can be writing by this way.

dohooo:
@IanVS updated.

packages/builder-vite/README.md
```diff
@@ -104,6 +104,33 @@ module.exports = {
...
+    },
+}
+
+export default config;
+```
+
+or
+
+```typescript
+import { StorybookViteConfig, ViteFinal } from '@storybook/builder-vite';
```
joshwooding, line 127:
```suggestion
import { ViteFinal } from '@storybook/builder-vite';
```
A minor thing

joshwooding:
Thanks @dohooo 🎉

Issue #345, opened 2022-04-21
[Bug] Rollup plugin working locally but not on static build


troyvassalotti:
### Describe the bug

We use Storybook with Lit. In our Lit components, we `import` SCSS files.

```ts
import { html, LitElement } from "lit";
import style from "./quote-block.scss";

@customElement("component-name")
export class ComponentName extends LitElement {
  static get styles() {
    return [ style ];
  }
}
...
```

To import SCSS and not need to wrap all our imported styles in the `unsafeCSS()` function from Lit, we use a plugin in our Vite config - [`rollup-plugin-postcss-lit`](https://github.com/umbopepato/rollup-plugin-postcss-lit). I've also added this plugin to `main.js` in Storybook.

![image](https://user-images.githubusercontent.com/45142012/164543291-b8fd21b3-95f2-44a9-a37a-f111e71ab66f.png)

Components render as expected with `start-storybook`, but the components do not render when using `build-storybook` and viewing the resulting static pages. We are met with console errors:

![image](https://user-images.githubusercontent.com/45142012/164542796-1bfc8a97-5e68-42a8-8fd2-352f2efea5b9.png)

### Environment

 - Node.js version: [e.g. v14.19.1]
 - NPM version: [e.g. 6.14.16]

### Additional context

Because the plugin works with a dev server, and after not finding an answer in other repos, I am hoping someone here can provide some insight into why this would be happening, thank you.

IanVS:
Thanks for the report.  I haven't worked with lit before, so I may not be a ton of help here, but I have a few questions.  Have you been using storybook with lit previously, and it started to break when you changed to the vite builder?  Does the rest of your app work just fine when building for production (without storybook)?  Are you using https://www.npmjs.com/package/@storybook/lit?  I don't see a `framework` specified in your `main.js` file.

We don't currently have an example project in this repo using lit.  It would be super useful if you were able to submit a PR to add one, ideally one which reproduces the problem you're having.  Then we'll be able to explore a bit and see if we can figure out what's going on.

IanVS:
Have you tried adding `?inline` to your imports? https://github.com/umbopepato/rollup-plugin-postcss-lit/issues/40#issuecomment-1063309922

troyvassalotti:
Hey @IanVS! Thanks for the quick response. I can clarify a bit on this:

[...]

PR #350, opened 2022-04-22
Add example for Lit TS

troyvassalotti:
# This PR has bugs

This PR adds an example for Lit TS to the examples directory. It is not extensive and doubles as a working example of #345 .

## To Reproduce Bug

- Run `yarn storybook` and see that the Simple Greeting component has blue text.
- Run `yarn build-storybook` and `yarn preview-storybook` to see that the component text did not accept the SASS styles from step 1.


examples/lit-ts/.storybook/main.ts
```diff
@@ -0,0 +1,16 @@
...
+module.exports = {
+  stories: ['../stories/**/*.stories.mdx', '../stories/**/*.stories.@(js|jsx|ts|tsx)'],
+  addons: ['@storybook/addon-links', '@storybook/addon-essentials'],
+  core: {
+    builder: '@storybook/builder-vite',
+  },
+  framework: '@storybook/web-components',
+  async viteFinal(config, { configType }) {
+    return mergeConfig(config, {
+      plugins: [postcssLit()],
```
IanVS, line 13: So, it seems that the default `include` from this plugin (`**/*.{css,sss,pcss,styl,stylus,sass,scss,less}`) is breaking with vite, because vite is adding query params to the id.  So, for example, the stylesheet in this PR is becoming `builder-vite/examples/lit-ts/stories/simplegreeting.scss?used`, which is filtered out.  I verified that updating the config as shown below fixes the issue:

```ts
// prettier-ignore
plugins: [postcssLit({ include: ['**/*.scss', '**/*.scss\?*'] })],
```

I had to add the ignore because prettier tries to remove the escape for the `?`.

I think this might be worth opening an issue in https://github.com/umbopepato/rollup-plugin-postcss-lit, if you're up for it.  In the meantime, if you'd like to make the change I suggested above, I'll be happy to move forward with adding this example to our project.

examples/lit-ts/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "preview-storybook": "http-server storybook-static --port 6005 --silent",
+    "test": "wait-on tcp:6005 && test-storybook --url 'http://localhost:6005'",
+    "test-ci": "run-p --race test preview-storybook"
+  },
+  "author": "",
+  "license": "MIT",
+  "devDependencies": {
+    "@babel/core": "^7.17.9",
+    "@mdx-js/react": "^1.6.22",
+    "@storybook/addon-actions": "^6.4.22",
```
IanVS, line 23: Can you please update these to the same versions used in the other examples?  If they get out-of-sync, multiple copies of various storybook modules get installed in the monorepo, and it causes a mess.

IanVS, line 23: Can you please update these to the same versions used in the other examples?  If they get out-of-sync, multiple copies of various storybook modules get installed in the monorepo, and it causes a mess.

examples/lit-ts/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "compilerOptions": {
+    "lib": ["es2019", "es2020.promise", "es2020.bigint", "es2020.string"],
+    "module": "CommonJS",
+    "target": "ES2019",
+    "declaration": true,
+    "sourceMap": true,
+    "inlineSources": true,
+    "skipLibCheck": true
+  }
```
IanVS, line 10: Would you mind adding `"experimentalDecorators": true` to avoid warnings?

IanVS, line 10: Would you mind adding `"experimentalDecorators": true` to avoid warnings?

examples/lit-ts/stories/SimpleGreeting.ts
```diff
@@ -0,0 +1,15 @@
+import { html, LitElement } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import style from './simplegreeting.scss';
```
IanVS, line 3: I'm not sure why, but I'm getting a typescript error in my editor on this line.  Are you seeing the same?

IanVS, line 3: I'm not sure why, but I'm getting a typescript error in my editor on this line.  Are you seeing the same?

IanVS (commented):
This is excellent, thanks so much for opening this PR.  I think I found the issue, mentioned in a comment below, and I had some small nits before we can merge this, but I'd like to get it in!

examples/lit-ts/.storybook/main.ts
troyvassalotti: Thanks for figuring that out, and so quickly too! I can confirm this worked for us and I've opened an issue in the plugin repo about it.

troyvassalotti:
When I get a minute I'll make those edits and see if we can this merged in. Thanks for helping figure out the underlying issue we were having!

troyvassalotti:
@IanVS Updates made - let me know if there's anything else I can do here!

IanVS:
@troyvassalotti thanks.  It looks like tests are failing.  Can you try building the examples locally and poking around at them using `npx http-serve storybook-static` (in the example directory)?  I wonder if the lit example didn't build correctly.

IanVS:
So close!  I'll try to take a look at the windows failures at some point, if you're not able to get to it.

troyvassalotti:
That's so strange - I'm even working on Windows and it worked for me running `yarn test-ci` in the `lit-ts` directory. Wonder what that could be?

IanVS:
It also passes just fine for me in a VM.  I've got no idea why it would fail in github.  I'm tempted to say we should just merge it and we can keep an eye on it for a while, and maybe it'll resolve, or we can dig into the failure later.

troyvassalotti:
I could add that `--passWithNoTests` flag to this example's test script as a workaround for the Windows test here not catching any stories? The macOS and Ubuntu tests should in theory ignore that flag since the tests are working as expected on both those.

IanVS:
Yeah, I think that's probably a good thing to do for now, until/unless we start hearing about windows problems from other uses.  Thanks!

IanVS:
I'm going to merge this as-is, because I'm updating some deps and want to include the ones in this example too.  I'll add the passWithNoTests in another PR.  Thanks for the contribution, @troyvassalotti!

PR #363, opened 2022-04-29
Warn in console when storyStoreV7 used with .svelte files

IanVS:
This makes it a bit more obvious that something is wrong when using storyStoreV7 with `.svelte` files (ref https://github.com/storybookjs/builder-vite/issues/361), and gives an indication of how to fix the issue:

image

This also fixes https://github.com/storybookjs/builder-vite/issues/356 by expanding the range for the @storybook/core-common peer dependency to avoid installing an older version when the user is using a 6.5 pre-release.

### DIFF
--- a/examples/svelte/.storybook/main.js
+++ b/examples/svelte/.storybook/main.js
@@ -7,6 +7,10 @@ module.exports = {
   core: {
     builder: '@storybook/builder-vite',
   },
+  features: {
+    // On-demand store does not work for .svelte stories, only CSF.
+    storyStoreV7: false,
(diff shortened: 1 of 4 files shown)


joshwooding (approved):

joshwooding:
This kind of feels like a feature that should be upstream but I don't see any harm adding it here for now.

IanVS:
Yeah I wanted it here because the cli defaults vite projects to storyStoreV7, and also I wanted a better warning when using 6.4 and not just 6.5.

PR #364, opened 2022-04-30
Prevent vite:define from replacing in story source

IanVS:
This fixes an issue where the vite:define plugin replaced environment variables inside the injected story source from `@storybook/source-loader`, which is supposed to be a string.  But when vite does the replacement, it adds its own quotes, which breaks the string.  

For a simplified example, storybook might generate something like this:
```js
var __STORY__ = "console.log(import.meta.env.VITE_VAR);"
```
But then vite:define turns it into:
```js
var __STORY__ = "console.log("env variable value");"
```

Here's a reproduction from the user who originally reported this issue: https://github.com/victorlmneves/vite-storybook

To test, I modified our environment variable stories.  In the process, I used some zero-width spaces to make the headings look a bit better, as suggested in https://vitejs.dev/guide/env-and-mode.html#production-replacement.

I also decided to use `ts-node` instead of adding a babel config for the react-ts project.  At least until we figure out exactly how that .babelrc works.

### DIFF
--- a/packages/builder-vite/source-loader-plugin.ts
+++ b/packages/builder-vite/source-loader-plugin.ts
@@ -1,21 +1,88 @@
 import sourceLoaderTransform from '@storybook/source-loader';
+import type { ExtendedOptions } from './types';
 
-export function sourceLoaderPlugin() {
-  return {
-    name: 'storybook-vite-source-loader-plugin',
-    enforce: 'pre',
-    async transform(src: string, id: string) {
-      if (id.match(/\.stories\.[jt]sx?$/)) {
-        // We need to mock 'this' when calling transform from @storybook/source-loader
-        // noinspection JSUnusedGlobalSymbols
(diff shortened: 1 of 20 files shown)


packages/builder-vite/source-loader-plugin.ts
```diff
@@ -1,21 +1,88 @@
...
+  return [
+    {
+      name: 'storybook-vite-source-loader-plugin',
+      enforce: 'pre',
+      buildStart() {
+        storySources.set(config, new Map());
+      },
+      async transform(src: string, id: string) {
+        if (id.match(storyPattern)) {
+          const code = await sourceLoaderTransform.call(mockClassLoader(id), src);
```
joshwooding, line 45: There are some hidden type issues but nothing that breaks functionality.

Because this returns any, sourceString below is any even though the type is strictly string | null which will upset replaceAll but the any means this isn't thrown.

joshwooding (approved):

joshwooding (approved):
Nice 👍🏽

PR #372, opened 2022-05-08
Disable HMR for web-components projects

### LINKED ISSUE #370: [Bug] storyStoreV7 is not compatible with framework: "@storybook/web-components"

2.9.7

Environment Info:

  System:
    OS: macOS 10.15.7
    CPU: (12) x64 Intel(R) Core(TM) i7-8700B CPU @ 3.20GHz
  Binaries:
    Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
    Yarn: 3.2.0 - ~/.nvm/versions/node/v16.14.0/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v16.14.0/bin/npm
  Browsers:
    Chrome: 101.0.4951.54
    Firefox: 100.0
    Safari: 15.0
  npmPackages:
    @storybook/addon-actions: ^6.4.22 => 6.4.22 
    @storybook/addon-essentials: ^6.4.22 => 6.4.22 
    @storybook/addon-links: ^6.4.22 => 6.4.22 
    @storybook/builder-vite: ^0.1.33 => 0.1.33 
    @storybook/web-components: ^6.4.22 => 6.4.22 

##  when not set storyStoreV7:

update /src/my-element.ts ,  got the Log : `[vite] page reload src/my-element.ts `

The expected result should be: `[vite] hmr update /src/my-element.ts`

## when set `storyStoreV7: true`,

update /src/my-element.ts, then got the error Log:

``` bash
Failed to execute 'define' on 'CustomElementRegistry': the name "my-element" has already been used with this registry
Error: Failed to execute 'define' on 'CustomElementRegistry': the name "my-element" has already been used with this registry
    at legacyCustomElement (http://localhost:6008/node_modules/.vite-storybook/deps/lit_decorators_js.js?v=53feff7d:16:25)
    at http://localhost:6008/node_modules/.vite-storybook/deps/lit_decorators_js.js?v=53feff7d:29:99
    at __decorateClass (http://localhost:6008/src/my-element.ts?t=1651908414168:7:57)
    at http://localhost:6008/src/my-element.ts?t=1651908414168:50:13
```

The expected result should be: `[vite] hmr update /src/my-element.ts `

https://github.com/boeto/test-vite-lit-storybook-story-store-V7

### Participation

- [ ] I am willing to submit a pull request for this issue.

IanVS:
Fixes #370.

When using `storyStoreV7`, we configure HMR to swap out the `importFn` when stories change.  However, this breaks web-components projects (see https://github.com/storybookjs/storybook/tree/master/app/web-components#setup-page-reload-via-hmr).  

So, instead, this change checks if the framework is `web-components` and if so, refreshes the page instead of performing an HMR.  It's still pretty quick, and not a full page reload.

You can test this out by starting up our `lit-ts` example (which I've configured to use storyStoreV7 now), open the story, and save the component or story file.  It shouldn't blow up anymore.  

One bummer is that MDX files also no longer HMR, though technically they could.  But I can't find a way to detect which part of the `importFn` changed, in order to determine whether to HMR or not, and it's safer to just avoid it for all stories when using web-components.

### DIFF
--- a/packages/builder-vite/codegen-modern-iframe-script.ts
+++ b/packages/builder-vite/codegen-modern-iframe-script.ts
@@ -17,11 +17,7 @@ export async function generateModernIframeScriptCode(options: ExtendedOptions) {
     if (framework === 'web-components') {
       return `
       if (import.meta.hot) {
-        import.meta.hot.accept((newModule) => {
-          const currentLocationHref = window.location.href;
-          window.history.pushState(null, null, currentLocationHref);
-          window.location.reload();
-        });
+        import.meta.hot.decline();
       }`.trim();
     }
 
--- a/packages/builder-vite/codegen-modern-iframe-script.ts
+++ b/packages/builder-vite/codegen-modern-iframe-script.ts
@@ -4,7 +4,7 @@ import { transformAbsPath } from './utils/transform-abs-path';
 import type { ExtendedOptions } from './types';
 
 export async function generateModernIframeScriptCode(options: ExtendedOptions) {
-  const { presets, configDir } = options;
+  const { presets, configDir, framework } = options;
 
   const previewOrConfigFile = loadPreviewOrConfigFile({ configDir });
   const presetEntries = await presets.apply('config', [], options);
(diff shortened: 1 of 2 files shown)


packages/builder-vite/codegen-modern-iframe-script.ts
```diff
@@ -4,22 +4,49 @@ import { transformAbsPath } from './utils/transform-abs-path';
...
     .map((configEntry) => transformAbsPath(configEntry));
 
-  // noinspection UnnecessaryLocalVariableJS
+  const generateHMRHandler = (framework: string): string => {
+    // Web components are not compatible with HMR, so disable HMR, reload page instead.
+    if (framework === 'web-components') {
+      return `
+      if (import.meta.hot) {
+        import.meta.hot.accept((newModule) => {
+          const currentLocationHref = window.location.href;
```
joshwooding, line 21: Can we change this to `import.meta.hot.decline()` ? https://vitejs.dev/guide/api-hmr.html#hot-decline

IanVS: 👍 I did a little benchmarking, and using `decline` is indeed a little faster.  It was consistently taking 0.75 sec, vs 0.7 - 1.0 the other way I had it.

joshwooding (approved):

PR #378, opened 2022-05-15
Fix order of MDX stories in sidebar

### LINKED ISSUE #264: Sub-stories are order alphabetical by default when using mdx
By default, stories in a folder should be ordered by there definition order.
Instead they are sorted alphabetically.
See e.g. here: https://github.com/bodograumann/storybook-jest-vite-interop, where `Button 1 / ZZZ` is defined before `Button 1 / AAA`, but the order in the storybook is reversed.
Screenshot_20220228-092303

joshwooding:
Closes #264


IanVS (approved):
Oh wild, we just weren't applying the export order plugin to MDX files.  Good find.

PR #379, opened 2022-05-15
Fix new stories not appearing in the importFn

### LINKED ISSUE #354: [Bug] New story files are not injected with storyStoreV7

2.9.1

```
System:
    OS: macOS 12.3.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  Binaries:
    Node: 16.13.1 - ~/.volta/tools/image/node/16.13.1/bin/node
    Yarn: 1.22.17 - ~/.volta/bin/yarn
    npm: 8.3.0 - ~/.volta/tools/image/npm/8.3.0/bin/npm
  Browsers:
    Firefox: 99.0.1
    Safari: 15.4
  npmPackages:
    @storybook/addon-a11y: 6.5.0-alpha.49 => 6.5.0-alpha.49
    @storybook/addon-actions: 6.5.0-alpha.49 => 6.5.0-alpha.49
    @storybook/addon-docs: 6.5.0-alpha.49 => 6.5.0-alpha.49
    @storybook/addon-essentials: 6.5.0-alpha.49 => 6.5.0-alpha.49
    @storybook/addon-links: 6.5.0-alpha.49 => 6.5.0-alpha.49
    @storybook/addon-postcss: ^2.0.0 => 2.0.0
    @storybook/addons: 6.5.0-alpha.49 => 6.5.0-alpha.49
    @storybook/builder-vite: ^0.1.27 => 0.1.27
    @storybook/expect: ^27.5.2-0 => 27.5.2-0
    @storybook/jest: ^0.0.10 => 0.0.10
    @storybook/react: 6.5.0-alpha.49 => 6.5.0-alpha.49
    @storybook/test-runner: ^0.0.5-next.2 => 0.0.5-next.2
    @storybook/testing-library: ^0.0.9 => 0.0.9
    @storybook/testing-react: ^1.2.4 => 1.2.4
```

### Describe the Bug

When `storyStoreV7` is enabled, adding a new story file will cause the story to appear in the sidebar correctly, but trying to visit that story will cause an error, because the `importFn` for the new story file is not added.

_No response_

### Participation

- [X] I am willing to submit a pull request for this issue.

joshwooding:
Fixes #354

### DIFF
--- a/packages/builder-vite/code-generator-plugin.ts
+++ b/packages/builder-vite/code-generator-plugin.ts
@@ -24,15 +24,24 @@ export function codeGeneratorPlugin(options: ExtendedOptions): Plugin {
     configureServer(server) {
       // invalidate the whole vite-app.js script on every file change.
       // (this might be a little too aggressive?)
-      server.watcher.on('change', (_e) => {
-        const { moduleGraph } = server;
-        const appModule = moduleGraph.getModuleById(virtualFileId);
+      server.watcher.on('change', () => {
+        const appModule = server.moduleGraph.getModuleById(virtualFileId);
         if (appModule) {
           server.moduleGraph.invalidateModule(appModule);
         }
-        const storiesModule = moduleGraph.getModuleById(virtualStoriesFile);
+        const storiesModule = server.moduleGraph.getModuleById(virtualStoriesFile);
         if (storiesModule) {
           server.moduleGraph.invalidateModule(storiesModule);
+        }
+      });
+
+      // Adding new story files is not covered by the change event above. So we need to detect this and trigger
+      // HMR to update the importFn.
+      server.watcher.on('add', (path) => {
+        // TODO maybe use the stories declaration in main
+        if (/\.stories\.([tj])sx?$/.test(path) || /\.(story|stories).mdx$/.test(path)) {
+          // We need to emit a change event to trigger HMR
+          server.watcher.emit('change', virtualStoriesFile);
         }
       });
     },


packages/builder-vite/code-generator-plugin.ts
```diff
@@ -25,16 +25,21 @@ export function codeGeneratorPlugin(options: ExtendedOptions): Plugin {
...
         }
-        const storiesModule = moduleGraph.getModuleById(virtualStoriesFile);
+        const storiesModule = server.moduleGraph.getModuleById(virtualStoriesFile);
         if (storiesModule) {
           server.moduleGraph.invalidateModule(storiesModule);
         }
       });
+
+      server.watcher.on('all', (_e, path) => {
+        if (/\.stories\.([tj])sx?$/.test(path) || /.mdx$/.test(path)) {
```
IanVS, line 39: should this be limited to `stories|story.mdx`?

joshwooding: Any mdx could be shown I thought? Maybe we should re-use the stories regex though

IanVS: I don't think anything other than a stories mdx will be adding new stories to the sidebar though, which I thought was the reason for this fix?

IanVS: And yeah, I think using the stories regex would be best here, if it works.  I guess we wouldn't be able to if they're using the object-style configuration though, right?  Or is there a utility to turn that into a normal glob pattern?

joshwooding: > I don't think anything other than a stories mdx will be adding new stories to the sidebar though, which I thought was the reason for this fix?

ah this just clicked. Good point

joshwooding: I'll just stick with regex for now to keep it simple

packages/builder-vite/code-generator-plugin.ts
```diff
@@ -25,16 +25,21 @@ export function codeGeneratorPlugin(options: ExtendedOptions): Plugin {
...
           server.moduleGraph.invalidateModule(appModule);
         }
-        const storiesModule = moduleGraph.getModuleById(virtualStoriesFile);
+        const storiesModule = server.moduleGraph.getModuleById(virtualStoriesFile);
         if (storiesModule) {
           server.moduleGraph.invalidateModule(storiesModule);
         }
       });
+
+      server.watcher.on('all', (_e, path) => {
```
IanVS, line 38: Maybe a comment about why this is being added would be good to have, for anyone else coming across this code later?

IanVS (approved):

IanVS (approved):
:+1:

PR #380, opened 2022-05-17
WIP: Support MDX2

IanVS:
This is an alternate approach to https://github.com/storybookjs/builder-vite/pull/377 that attempts to add support for MDX2 (https://github.com/storybookjs/builder-vite/issues/234).

This applies a similar approach as [vite-plugin-mdx](https://github.com/brillout/vite-plugin-mdx), sending the output of `@storybook/mdx2-csf` through the `@vitejs/plugin-babel` transform in order to add react refresh support.  This currently blows up because `@storybook/mdx2-csf` processes the code with babel, which adds its own react-transform.  @shilman has indicated that it shouldn't be a problem to add a method or flag that would expose the src without the babel transforms.

### DIFF
--- a/packages/builder-vite/mdx-plugin.ts
+++ b/packages/builder-vite/mdx-plugin.ts
@@ -1,5 +1,14 @@
+// import { ExtendedOptions } from '../types';
 import mdx from 'vite-plugin-mdx';
-import { createCompiler } from '@storybook/csf-tools/mdx';
+import { createFilter } from '@rollup/pluginutils';
+import { Plugin } from 'vite';
+const { logger } = require('@storybook/node-logger');
+const { createCompiler } = require('@storybook/csf-tools/mdx');
+
+type Options = {
+  include?: string | RegExp | (string | RegExp)[];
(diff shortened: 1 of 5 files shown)


packages/builder-vite/mdx-plugin.ts
```diff
@@ -9,7 +18,50 @@ import { createCompiler } from '@storybook/csf-tools/mdx';
...
+            p.name === 'react-refresh' ||
+            p.name === 'vite:react-babel' ||
+            p.name === 'vite:react-refresh' ||
+            p.name === 'vite:react-jsx'
+        );
+        reactRefreshPlugin = reactRefreshPlugins.find((p: any) => p.transform);
+      },
+      async transform(src: string, id: string, ssr) {
+        if (!/\.mdx?$/.test(id)) return src;
+        const code = await compile(src, { skipCsf: true });
```
IanVS, line 49: We won't want to `skipCsf` in the final version, but it avoids some errors for now.

IanVS:
Replaced by https://github.com/storybookjs/builder-vite/pull/377

PR #382, opened 2022-05-18
Handle new versions of vite-plugin-svelte

IanVS:
Fixes https://github.com/storybookjs/builder-vite/issues/381

There have been two changes in @sveltejs/vite-plugin-svelte that cause problems for us, which this PR addresses:

- https://github.com/sveltejs/vite-plugin-svelte/releases/tag/%40sveltejs%2Fvite-plugin-svelte%401.0.0-next.42 started to merge the `svelte.config.js` file [by default](https://github.com/sveltejs/vite-plugin-svelte/pull/317), which may not be what storybook users want / expect, since their production config may be different from the storybook config (same reason we don't automatically merge `vite.config.js`).  With the changes here, the `configFile` option in `svelteOptions` can still be set to `true` or to a specific file, if desired.

- https://github.com/sveltejs/vite-plugin-svelte/releases/tag/%40sveltejs%2Fvite-plugin-svelte%401.0.0-next.43 began [returning an array of plugins](https://github.com/sveltejs/vite-plugin-svelte/pull/322) rather than just a single one.  This would normally be fine, except that we hack around with the plugin to create two separate ones, with different names, to avoid HMR in story files.  So, this PR checks to see if the result is an array or not, and reacts accordingly.  Note, I'm throwing away the inspector for story files.  This might be a problem, but until we hear about it, I think this'll be fine.  The inspector is still pretty experimental anyway, from what I can tell.

To test, you can revert the changes I've made in `vite-config.ts`, and you'll see that the `svelte` example breaks (I updated the svelte plugin in it).  With the changes, it builds successfully.

This also enables the use of the new experimental inspection plugin.

image

To enable this feature, be sure you have `@sveltejs/vite-plugin-svelte` version `1.0.0-next.43` or higher, then add the following to your `svelteOptions` in `.storybook/main.js`:

```js
module.exports = {
  // ... rest of config
  framework: '@storybook/svelte',
  svelteOptions: {
    preprocess: preprocess(),
    experimental: { inspector: true }, // <-- this line
  },
};
```

Then, when the storybook opens, focus a story iframe and press cmd+shift (mac) or ctrl+shift (windows) to activate.

### DIFF
--- a/packages/builder-vite/vite-config.ts
+++ b/packages/builder-vite/vite-config.ts
@@ -97,10 +97,20 @@ export async function pluginConfig(options: ExtendedOptions, _type: PluginConfig
       // These are the svelte stories we need to exclude from HMR
       const storyPatterns = ['**/*.story.svelte', '**/*.stories.svelte'];
       // Non-story svelte files
-      plugins.push(sveltePlugin({ ...svelteOptions, exclude: [...userExclude, ...storyPatterns] }));
+      // Starting in 1.0.0-next.42, svelte.config.js is included by default.
+      // We disable that, but allow it to be overridden in svelteOptions
+      plugins.push(sveltePlugin({ configFile: false, ...svelteOptions, exclude: [...userExclude, ...storyPatterns] }));
       // Svelte stories without HMR
+      const storySveltePlugin = sveltePlugin({
+        configFile: false,
+        ...svelteOptions,
+        exclude: userExclude,
+        include: storyPatterns,
+        hot: false,
+      });
       plugins.push({
-        ...sveltePlugin({ ...svelteOptions, exclude: userExclude, include: storyPatterns, hot: false }),
+        // Starting in 1.0.0-next.43, the plugin function returns an array of plugins.  We only want the first one here.
+        ...(Array.isArray(storySveltePlugin) ? storySveltePlugin[0] : storySveltePlugin),
         name: 'vite-plugin-svelte-stories',
       });
     } catch (err) {
--- a/examples/svelte/.storybook/main.js
+++ b/examples/svelte/.storybook/main.js
@@ -17,5 +17,8 @@ module.exports = {
   },
   svelteOptions: {
     preprocess: preprocess(),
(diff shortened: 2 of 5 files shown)


packages/builder-vite/vite-config.ts
```diff
@@ -97,10 +97,20 @@ export async function pluginConfig(options: ExtendedOptions, _type: PluginConfig
...
+        configFile: false,
+        ...svelteOptions,
+        exclude: userExclude,
+        include: storyPatterns,
+        hot: false,
+      });
       plugins.push({
-        ...sveltePlugin({ ...svelteOptions, exclude: userExclude, include: storyPatterns, hot: false }),
+        // Starting in 1.0.0-next.43, the plugin function returns an array of plugins.  We only want the first one here.
+        ...(Array.isArray(storySveltePlugin) ? storySveltePlugin[0] : storySveltePlugin),
```
joshwooding, line 113: Nice, my first question was going to be if this breaks previous versions but I can see that it shouldn't :P

packages/builder-vite/vite-config.ts
```diff
@@ -97,10 +97,20 @@ export async function pluginConfig(options: ExtendedOptions, _type: PluginConfig
       // These are the svelte stories we need to exclude from HMR
       const storyPatterns = ['**/*.story.svelte', '**/*.stories.svelte'];
       // Non-story svelte files
-      plugins.push(sveltePlugin({ ...svelteOptions, exclude: [...userExclude, ...storyPatterns] }));
+      // Starting in 1.0.0-next.42, svelte.config.js is included by default.
+      // We disable that, but allow it to be overridden in svelteOptions
+      plugins.push(sveltePlugin({ configFile: false, ...svelteOptions, exclude: [...userExclude, ...storyPatterns] }));
```
joshwooding, line 102: Does this also need to only use the first plugin if it's an array?

IanVS: No, the reason we need to for the other one is that we're expecting a single object, spreading the values, and changing the `name` field.  Vite itself is fine with arrays of plugins, it flattens them all out at the end.

joshwooding (approved):

PR #394, opened 2022-05-26
Support Vite 3

IanVS:
This updates the vite builder to support vite 3.0, which is a breaking change due to our use of @vitejs/plugin-react, which also has a new major version to support vite 3.  

This also updates our examples to the latest vite and vite plugin beta versions.


IanVS:
We're getting an error from `memoizerific`, which is used in many storybook packages, and has a check for `process.env.FORCE_SIMILAR_INSTEAD_OF_MAP`.  Why was this working before? 🤔

IanVS:
Cool, after rebasing on to https://github.com/storybookjs/builder-vite/pull/410, the examples built successfully in the linux runner.  The mac build ran out of memory, but otherwise looked like it would have succeeded.

IanVS:
Building the examples was running out of memory in the macOS runner, and my theory is that it is due to those runners having three cores, and yarn choosing to run two concurrent builds in parallel, which exceeds the memory available to node on those worker boxes.  But they have 14gb of memory available, so I'm trying to bump the `--max-old-space-size`  to see if that resolves the issue.

PR #399, opened 2022-05-31
Allow setting global args in SB 6.5+

IanVS:
Fixes https://github.com/storybookjs/builder-vite/issues/395.

This is a little bit tricky to do in a way that supports storybook 6.4 and 6.5.  I've used a dynamic `import()` and a top-level await, in order to wrap the import of the new methods in a try/catch.  This has _decent_ [browser support](https://caniuse.com/mdn-javascript_operators_await_top_level), but is not universal.  Curious if anyone else has another idea.

I've added example global args in the `react` and `react-ts` examples, to show both story store V6 and V7 support.  You can simulate storybook 6.4 by changing the destructured imports to `addArgsZ` and `addArgTypesZ`, building, and starting an example.  You shouldn't see the control, and there shouldn't be any console errors.

### DIFF
--- a/packages/builder-vite/codegen-iframe-script.ts
+++ b/packages/builder-vite/codegen-iframe-script.ts
@@ -37,13 +37,30 @@ export async function generateIframeScriptCode(options: ExtendedOptions) {
 
     const configs = [${importArray('config', configEntries.length).concat('preview.default').join(',')}].filter(Boolean)
 
+    try {
+      // these were added in 6.5, but we support 6.4 as well.
+      var {addArgs, addArgTypes} = await import('@storybook/client-api');
+    } catch (err) {
+      // ignore error
+    }
+
     configs.forEach(config => {
-      Object.keys(config).forEach((key) => {
+      Object.keys(config).forEach(async (key) => {
         const value = config[key];
         switch (key) {
-          case 'args':
+          case 'args': {
+            if (typeof addArgs !== 'undefined') {
+              return addArgs(value);
+            } else {
+              return logger.warn('Could not add global args. Please open an issue in storybookjs/builder-vite.');
+            }
+          }
           case 'argTypes': {
-            return logger.warn('Invalid args/argTypes in config, ignoring.', JSON.stringify(value));
+            if (typeof addArgTypes !== 'undefined') {
+              return addArgTypes(value);
+            } else {
+              return logger.warn('Could not add global argTypes. Please open an issue in storybookjs/builder-vite.');
+            }
           }
           case 'decorators': {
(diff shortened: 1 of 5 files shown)


IanVS:
We got failures in some of the example builds (strangely, not all of them).

>ERROR: Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2019", "firefox78", "safari13.1")

Should we update our browser support?  Or, maybe I need to wrap the whole thing in an async IIFE?

joshwooding:
Hmm 86% supported globally seems fine to me. I don't have a strong preference either way

joshwooding:
Would something like this work:

```js
 import * as clientApi from "@storybook/client-api";
const {
  addDecorator,
  addParameters,
  addLoader,
  addArgTypesEnhancer,
  addArgsEnhancer,
  setGlobalRender,
} = clientApi;

  case "args": {
    if (typeof clientApi.addArgs !== "undefined") {
      return clientApi.addArgs(value);
    } else {
      return logger.warn(
        "Could not add global args. Please open an issue in storybookjs/builder-vite."
      );
    }
  }
  case "argTypes": {
    if (typeof clientApi.addArgTypes !== "undefined") {
      return clientApi.addArgTypes(value);
    } else {
      return logger.warn(
        "Could not add global argTypes. Please open an issue in storybookjs/builder-vite."
      );
    }
  }


```

IanVS:
Brilliant, that works great.  I tested with storybook 6.4 and 6.5, and it behaves correctly, although react is yelling about trying to set `globalArg` as a prop...

packages/builder-vite/codegen-iframe-script.ts
```diff
@@ -22,28 +22,44 @@ export async function generateIframeScriptCode(options: ExtendedOptions) {
...
-    import * as preview from '${virtualPreviewFile}';
-    import { configStories } from '${virtualStoriesFile}';
+      setGlobalRender,
+    } = clientApi;
 
     const configs = [${importArray('config', configEntries.length).concat('preview.default').join(',')}].filter(Boolean)
 
     configs.forEach(config => {
-      Object.keys(config).forEach((key) => {
+      Object.keys(config).forEach(async (key) => {
```
joshwooding, line 43:
```suggestion
      Object.keys(config).forEach((key) => {
```

joshwooding: We don't need this anymore :)

IanVS:
@joshwooding this is ready for another review whenever you've got some time.

joshwooding (approved):
LGTM

PR #404, opened 2022-06-03
Add composed "overview" example that builds/includes all framework examples

dohooo:
(description added by @ianvs)

This adds an example project (overview) which composes the framework examples together into a single storybook, which not only provides a demonstration of how this composition can be accomplished in a nice clean way, but also will allow us to use chromatic snapshots against all example storybooks at once.  

It also will publish this composed storybook to github pages, under the `/builder-vite` path.  It's being set up to only deploy to gh-pages on pushes to main.  But for now, you can see the result of the deploy in @dohooo's repo: https://dohooo.github.io/builder-vite/overview


examples/svelte/.storybook/main.ts
```diff
@@ -1,6 +1,7 @@
-const preprocess = require('svelte-preprocess');
+import { withOverview } from 'bookcase-builder';
+import preprocess from 'svelte-preprocess';
 
-module.exports = {
+export default withOverview(__dirname)({
```
IanVS, line 4: Did these need to change to TS files, or is that just for convenience?

examples/svelte/package.json
```diff
@@ -33,5 +33,6 @@
     "typescript": "^4.5.5",
     "vite": "2.9.6",
     "wait-on": "^6.0.1"
-  }
+  },
+  "bookcase-builder": {}
```
IanVS, line 37: Is this needed, if there's no configuration here?  I'd prefer if this empty config wasn't necessary.

package.json
```diff
@@ -35,14 +36,26 @@
...
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-only-warn": "^1.0.3",
     "eslint-plugin-storybook": "^0.5.5",
+    "gh-pages": "^3.2.3",
     "prettier": "^2.5.1",
     "typescript": "^4.5.4"
+  },
+  "dependencies": {
+    "bookcase-builder": "^0.0.8",
+    "types-package-json": "^2.0.39"
```
IanVS, line 51: Is types-package-json still required?

packages/builder-vite/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "strict": true,
     "rootDir": ".",
     "outDir": "dist",
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "types": []
```
IanVS, line 8: Why is this change needed?

package.json
```diff
@@ -8,6 +8,7 @@
   },
   "scripts": {
     "build-examples": "yarn workspaces foreach -v --include 'example-*' -p run build-storybook",
+    "build-examples:overview": "bb",
     "test-ci": "yarn workspaces foreach  -v --exclude 'example-svelte' -p run test-ci",
```
IanVS, line 12: We need to `--exclude 'example-overview'` since it has no stories of its own to test.  That's why CI is failing right now.

IanVS (commented):
This is looking nice and clean!  🧹

I had a few questions and a suggestion for how to get CI passing.

examples/svelte/.storybook/main.ts
dohooo: For good type hints, of course, you can use `.js`, and bookcase-builder supports cjs.

examples/svelte/package.json
dohooo: https://github.com/dohooo/bookcase-builder.

```shell
// package.json in the workspace directory.
// Here you can configure the project to be built. For example specify the storybook configuration directory.
// If you want to use the default configuration, you can set it to "{}". And if haven't configured it, BB will't build it.
{
  "bookcase-builder": {
    // Rename the current package name, default to package directory name.
    "basename": "other name",
    "storybook": {
      // Storybook configuration directory, default to ".storybook"
      "configDir": "./.otherStorybookDir"
    }
  }
}
```

IanVS: OK, I think it's a little confusing to have an empty config object that has meaning (causes it to be included). IMHO it would be more clear to have an explicit `include: true` (or maybe `exclude: true` if you don't want it).  Or maybe even better, a config only at the top level, with a list of subprojects to build, which would mean that each subproject does not need to have a config.  But, those are just suggestions.

examples/svelte/.storybook/main.ts
IanVS: OK, normally I would want somewhat-unrelated changes like that to be in a different PR.  This PR should really just be the smallest number of changes to make the composition work correctly.  And there are problems with using ts config files in some cases in storybook (e.g. when `"type": "module"` is set), so I'm not sure if it will confuse others if they try to use these examples as a base for their own projects.  But, we can try this, and change it later if there is confusion.

dohooo: Ok, I learned. 👍

IanVS (approved):
Great work here, I'm excited to try this out with chromatic so we can publish all our stories together, and get snapshots set up.

PR #410, opened 2022-06-15
Use an alias for react-dom/client if not installed

IanVS:
This implements a suggestion in https://github.com/vitejs/vite/issues/6007#issuecomment-1155720679
to alias react-dom/client to our placeholder file when the dependency is not found.

In Vite 3.0, our current approach will no longer work, and we'll need to use this method instead.  But it works in Vite 2 as well, so we might as well merge this now while we work out the rest of Vite 3 compatibility.


joshwooding (approved):
LGTM!

PR #413, opened 2022-06-20
Support missing import.meta.env

princed:
Currently vite storybook builder breaks when any env var used in code via `import.meta.env` is missing while building storybooks.

Apparently that happens because Rollup does direct inline replacements and turns code like

```ts
const getMyVar = () => import.meta.env.VITE_MISSING_VAR === 'true'
```

into something like

```js
const getMyVar = () => {"BASE_URL":"/","MODE":"production","DEV":false,"PROD":true}.VITE_MISSING_VAR === "true";
```

which is syntactically incorrect.
See my reproduction repo for an example — https://github.com/princed/vite-sb-meta-env-repro. 

To mitigate this `vite` itself [adds a replacement](https://github.com/vitejs/vite/blob/908c9e4cdd2cceb0f01495e38066ffe33c21ddb8/packages/vite/src/node/plugins/define.ts#L50) from `import.meta.env.` to `({}).` to catch missing direct replacements and produce correct code. I opted for identical solution here.

A similar workaround is also possible on the user side:

```js
async viteFinal(config) {
    config.define = {
      'import.meta.env.': `({}).`,
    };
   return config; 
}

### DIFF
--- a/packages/builder-vite/envs.ts
+++ b/packages/builder-vite/envs.ts
@@ -45,6 +45,9 @@ export function stringifyProcessEnvs(raw: EnvsRaw, envPrefix: UserConfig['envPre
   // support destructuring like
   // const { foo } = import.meta.env;
   envs['import.meta.env'] = JSON.stringify(stringifyEnvs(updatedRaw));
+  // Prevent build breaking on a missing var, similar to vite
+  // @see https://github.com/vitejs/vite/blob/908c9e4cdd2cceb0f01495e38066ffe33c21ddb8/packages/vite/src/node/plugins/define.ts#L50
+  envs['import.meta.env.'] = '({}).';
 
   return envs;
 }

### CODE REVIEW

File: packages/builder-vite/envs.ts [hunk 1]
```diff
@@ -22,7 +22,7 @@ export const allowedEnvPrefix = ['VITE_', 'STORYBOOK_'];
  * Customized version of stringifyProcessEnvs from @storybook/core-common which
  * uses import.meta.env instead of process.env and checks for allowed variables.
  */
-export function stringifyProcessEnvs(raw: EnvsRaw, envPrefix: UserConfig['envPrefix']) {
+export function stringifyProcessEnvs(raw: EnvsRaw, envPrefix: UserConfig['envPrefix'], isBuild: boolean) {
```
princed, line 25: Not sure what the preferred way to do this is, but happy to change it if needed.

princed:
> It seems like maybe the vite solution still has a kink to work out? https://github.com/vitejs/vite/issues/8663. Maybe we should hold off until that issue is solved, before adding this.

Happy to wait and see whether Vite team finds another approach to the problem, but it seems unlikely though.
We also could just merge this for now, and then handle changes (if any) when adding Vite 3 support (which will surely require some dedicated effort).

packages/builder-vite/envs.ts
IanVS: Let's use a `mode` of either `production` or `development`, to avoid a [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap).

princed: I feel like `mode` would be confusing here, because it's possible to build in `development` mode and serve in `production`. And `vite` itself uses `command` to derive this information: 
https://github.com/vitejs/vite/blob/908c9e4cdd2cceb0f01495e38066ffe33c21ddb8/packages/vite/src/node/plugins/define.ts#L12

princed: Would you be okay with `command: 'build' | 'serve'` as an argument?

IanVS: Sounds good, thanks.

princed: Done!

IanVS:
Vite 3 has been released, and https://github.com/vitejs/vite/issues/8663 is not yet solved.  I'm happy to proceed with this if the underlying problem is still an issue in Vite 3 (would you mind confirming?).

princed:
Don't mind at all, but will only be able to do that in a week.

princed:
So with Vite 3 and https://github.com/vitejs/vite/issues/8663 not fixed, even this PR doesn't fix building storybooks ¯\\\_(ツ)_/¯

PR #417, opened 2022-06-22
Add mdx preact example

IanVS:
This fixes an issue where as long as preact is installed in the monorepo, vite-plugin-mdx will default to using it, even in react projects, causing an error in mdx stories.

This uses the `framework` option, and only uses the preact mdx if the framework is preact, otherwise falling back to react.  This was enabled by https://github.com/brillout/vite-plugin-mdx/commit/b37e2cd168a0eaf844ed032513ceb85602b3cd5b.

I also added an mdx story in the preact example, but the styling is wrong.  @maxbeatty do you know why that might be?  Does storybook's emotion need to be configured to work with preact, maybe?

To test, verify that the `introduction` story is working in the react example as well as the preact example, and maybe vue too, for good measure.


joshwooding (approved):

maxbeatty:
Thanks for digging into this!

IanVS:
No problem @maxbeatty.  I could still use help figuring out why the styling isn't working correctly in the preact example mdx story.

PR #427, opened 2022-07-06
Add `.storybook` to `server.fs.allow`

benmccann:
Closes https://github.com/storybookjs/builder-vite/issues/426

### DIFF
--- a/packages/builder-vite/vite-config.ts
+++ b/packages/builder-vite/vite-config.ts
@@ -40,6 +40,11 @@ export async function commonConfig(
     cacheDir: 'node_modules/.vite-storybook',
     envPrefix,
     define: {},
+    server: {
+      fs: {
+        allow: ['.storybook']
+      }
+    },
     resolve:
       framework === 'vue3'
         ? {
--- a/packages/builder-vite/vite-config.ts
+++ b/packages/builder-vite/vite-config.ts
@@ -72,6 +72,10 @@ export async function pluginConfig(options: ExtendedOptions, _type: PluginConfig
       name: 'vite-plugin-storybook-allow',
       enforce: 'post',
       config(config) {
+        // if there is no allow list then Vite allows anything in the root directory
+        // if there is an allow list then Vite allows anything in the listed directories
+        // add the .storybook directory only if there's an allow list so that we don't end up
+        // disallowing the root directory unless it's already disallowed
         if (config?.server?.fs?.allow) {
           config.server.fs.allow.push('.storybook');
         }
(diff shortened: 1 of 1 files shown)


IanVS:
Thanks for the PR.  Do you know why this might be necessary?  With my understanding of the `fs.strict` config option (https://vitejs.dev/config/#server-fs-strict), files and folders inside the root should be allowed by default.  Is this maybe a bug in vite?

benmccann:
I think the docs are wrong. As soon as you specify one directory in the allow list it no longer allows the workspace root:

https://github.com/vitejs/vite/blob/f280dd99856af68a47635dda609c402333805209/packages/vite/src/node/server/index.ts#L683

SvelteKit adds directories to the allow list by default, which ends up removing the workspace root:

https://github.com/sveltejs/kit/blob/0915cb5c1a6edb07ac915c24c69cb47d1d15e3de/packages/kit/src/vite/index.js#L169

IanVS:
Oh I see, thanks.  In that case, I'm reluctant to merge this change, as it would have the same impact for everyone else not using sveltekit who might need files in other directories within the workspace root.  I don't suppose the sveltekit vite plugin could re-add the workspace root, so as to restore the default behavior?

packages/builder-vite/vite-config.ts
```diff
@@ -68,6 +68,15 @@ export async function pluginConfig(options: ExtendedOptions, _type: PluginConfig
       // Do not treat story files as HMR boundaries, storybook itself needs to handle them.
       exclude: [/\.stories\.([tj])sx?$/, /node_modules/].concat(framework === 'react' ? [] : [/\.([tj])sx?$/]),
     }),
+    {
+      name: 'vite-plugin-storybook-allow',
+      enforce: 'post',
+      config(config) {
+        if (config?.server?.fs?.allow) {
```
IanVS, line 75: Ok, so this will catch situations like the sveltekit plugin adding an allow list, which then prevents the root from being used.  I like this solution, but would you mind adding a quick comment somewhere here that explains why it's needed?  Otherwise, there's only about a 10% chance I'll remember why we're doing this in a few months.  :).

IanVS: Nevermind, I'll add the comment.  I want to get a release out with the MDX fix, and would be nice to include this as well.

benmccann: done

IanVS: Hah, you beat me to it!  Thanks.

IanVS (approved):

IanVS:
Thanks for the contribution!

PR #437, opened 2022-07-13
Improve sourcemaps

IanVS:
Turns out we weren't creating good sourcemaps, especially in dev.  It seems that whereas rollup only uses the `mappings`, I guess that vite/esbuild uses the `sources` field of the sourcemap as well, and without passing the `source` option to MagicString's `generateMap`, we were losing the connection.  So, this adds the `id` as the source, and also enables `hires` (high resolution) so that columns are correct and not just lines (useful for code coverage, for example).

I also added MagicString to some of our other plugins which weren't using it.

The most obvious changes are in the react and svelte projects.  Vue is a bit improved, but we're still losing the story file sourcemaps for some reason.

To test, fire up an example project (in dev especially, but good to check prod builds too), look in the sources tab in devtools, and poke around, looking for sourcemapped versions of the files.  You should see them in this branch, but not nearly as many, if any, in main.  

@yannbf, I also tested these changes by copying the `dist` folder from this project into the node_modules of your https://github.com/yannbf/storybook-coverage-recipes project, and got good results from the react_vite and svelte_vite examples there.

### DIFF
--- a/packages/builder-vite/svelte/csf-plugin.ts
+++ b/packages/builder-vite/svelte/csf-plugin.ts
@@ -4,6 +4,7 @@ import { extractStories } from '@storybook/addon-svelte-csf/dist/cjs/parser/extr
 const parser = require.resolve('@storybook/addon-svelte-csf/dist/esm/parser/collect-stories').replace(/[/\\]/g, '/');
 import type { Options } from '@sveltejs/vite-plugin-svelte';
 import * as svelte from 'svelte/compiler';
+import MagicString from 'magic-string';
 
 export default function csfPlugin(svelteOptions?: Options) {
   return {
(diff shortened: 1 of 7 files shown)


yannbf:
This is incredible. Thank you so much @IanVS!

packages/builder-vite/svelte/csf-plugin.ts
```diff
@@ -23,22 +25,26 @@ export default function csfPlugin(svelteOptions?: Options) {
...
-        const codeWithoutDefaultExport = code.replace('export default ', '// export default ');
+        s.replace('export default', '// export default');
 
         const namedExportsOrder = Object.entries<any>(stories)
           .filter(([, def]) => !def.template)
           .map(([id]) => id);
 
         const output = [
-          codeWithoutDefaultExport,
+          '',
```
joshwooding, line 35: We could just remove this right?

joshwooding (approved):

packages/builder-vite/svelte/csf-plugin.ts
IanVS: No, it's needed so that a newline is added.

PR #443, opened 2022-07-14
Ignore applyDecorators in v6 story store

kvcvc:
In https://github.com/storybookjs/storybook/commit/fcb61e10a3fb0ef5c59137c09c7766699a309b14 `decorateStory` was renamed by @yannbf to `applyDecorators` but the same author didn't think to rename potential references leading to a `applyDecorators was not supported :( !` log and I assume decorators not being applied when using Storybook. This pull request aims to solve the issue.

### DIFF
--- a/packages/builder-vite/codegen-iframe-script.ts
+++ b/packages/builder-vite/codegen-iframe-script.ts
@@ -85,6 +85,7 @@ export async function generateIframeScriptCode(options: ExtendedOptions) {
             v[key] = value;
             return addParameters(v, false);
           }
+          case 'decorateStory':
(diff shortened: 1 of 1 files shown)


IanVS:
Hi, the commit you linked to has not been released in any version of storybook outside of the 7.0 alphas.  How were you seeing that warning, since the vite builder is not yet supported in version 7?

kvcvc:
Hi, Ian, I believe the change was part of [6.5.9](https://www.npmjs.com/package/@storybook/vue3/v/6.5.9), or at least that's the version I experienced the issue with.

PR #444, opened 2022-07-14
cleanup dependencies on alphas/betas

benmccann:


package.json
```diff
@@ -47,7 +47,7 @@
     "typescript": "^4.5.4"
   },
   "dependencies": {
-    "@sveltejs/vite-plugin-svelte": "^1.0.0-next.49",
+    "@sveltejs/vite-plugin-svelte": "^1.0.0",
```
IanVS, line 50: I didn't catch this in https://github.com/storybookjs/builder-vite/pull/428/files, but I think this needs to move to the svelte example, right?

benmccann: I don't think it can. `builder-vite` uses `loadSvelteConfig` and so it needs `vite-plugin-svelte`

It sounds like it will become a non-issue with Storybook 7. If there were a need to clean it up before then we could use a bundler like Rollup and make it a devDependency so that it will just extract out `loadSvelteConfig` and bundle it with the project so that users don't end up downloading the full `vite-plugin-svelte`

IanVS: Oh right, gotcha.  Then it should move into the dependencies of `packages/builder-vite`, not the top level of the monorepo.

IanVS:
Regarding vite-plugin-svelte, I did make it load conditionally in https://github.com/storybookjs/builder-vite/pull/434, and if it's not installed we'll throw an error message: https://github.com/storybookjs/builder-vite/blob/067bda7762113b7b1e9a4e8a137ea0c40b8c42ef/packages/builder-vite/vite-config.ts#L138-L139

Let's remove it from dependencies after all, to avoid non-svelte projects from having to install it.

benmccann:
Ok. Updated!

IanVS:
Thanks!

PR #445, opened 2022-07-14
Cleanup `peerDependencies`

benmccann:
@storybook/source-loader doesn't need to be in peerDependencies because it's already in dependencies

@storybook/core-common and @storybook/node-logger can be just regular dependencies instead of peerDependencies since we don't need the user to add them to their package.json

~~@storybook/mdx1-csf doesn't need to be downloaded by all users. Users will use either @storybook/mdx1-csf or @storybook/mdx2-csf so move it to devDependencies so that users won't end up including it automatically~~

### DIFF
--- a/packages/builder-vite/package.json
+++ b/packages/builder-vite/package.json
@@ -14,7 +14,7 @@
   "homepage": "https://github.com/storybookjs/builder-vite/#readme",
   "dependencies": {
     "@joshwooding/vite-plugin-react-docgen-typescript": "0.0.4",
-    "@storybook/mdx1-csf": "^0.0.4",
+    "@storybook/node-logger": ">=6.4.3",
     "@storybook/source-loader": "^6.4.3",
     "@vitejs/plugin-react": "^2.0.0",
     "ast-types": "^0.14.2",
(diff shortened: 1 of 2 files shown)


packages/builder-vite/package.json
```diff
@@ -27,17 +27,15 @@
     "sveltedoc-parser": "^4.2.1"
   },
   "devDependencies": {
+    "@storybook/mdx1-csf": "^0.0.4",
```
IanVS, line 30: `@storybook/mdx1-csf` is included as a dependency of `@storybook/addon-docs`.  However, it uses `@storybook/mdx1-csf@^0.0.1`, but we require `^0.0.3` because of https://github.com/storybookjs/mdx1-csf/pull/9.

IanVS (commented):
Thanks for the cleanup.  The only one I don't think we can move is `@storybook/mdx1-csf`.  Since we require a particular version.

packages/builder-vite/package.json
benmccann: ok. reverted that

IanVS (approved):

PR #449, opened 2022-07-18
fix: pass options read from config file to svelteDocgen

benmccann:
Looks like we only passed the options from the config file to `csfPlugin` and missed `svelteDocgen`


IanVS (approved):

IanVS:
Thanks!

PR #457, opened 2022-07-23
Bump @joshwooding/vite-plugin-react-docgen-typescript to 0.0.5

joshwooding:


theshortcut:
thanks for this!

IanVS (approved):
@joshwooding what's the main change in this version?  I took a look, and I see that you're using the `@rollup/pluginutils` filter function, and building the plugin with a different tool.  The change to use the filter function is what fixes the problems we were seeing in monorepos, right?

joshwooding:
@IanVS yeah the filter change is what fixed it. The build tool was to move to the same tool that Vite now uses.

PR #461, opened 2022-07-25
Remove mock-core-js plugin

IanVS:
Fixes https://github.com/storybookjs/builder-vite/issues/143
Fixes https://github.com/storybookjs/builder-vite/issues/412
Fixes https://github.com/storybookjs/builder-vite/issues/448

I'm not sure why this plugin is here, but it seems to cause problems according to the issues mentioned here, and those problems are solved when this plugin is removed.  Maybe older versions of vite could not handle core-js?  @eirslett do you remember why you added this mock?

At any rate, since several folks have reported success when removing the plugin, let's just take it out.  If we hear it causes problems for others, we can investigate further what's going on.


joshwooding (approved):
This has been on my mind for a while! Happy to see it removed. If it causes any issues we can always add it back :)

princed:
I think just removing the plugin wasn't enough to solve the problem similar to one outlined in https://github.com/storybookjs/builder-vite/issues/143. In my case I also had to add `core-js-pure` to `optimizeDeps` in vite config.

eirslett:
I think the problem was that core-js code is in CommonJS format, so it had to be either preoptimized (every single import path from core-js, so it would be a very long list of imports) or mocked away.

If everything works as it should without mock-core-js then I'm sure it's better to just remove it.

PR #470, opened 2022-08-01
Remove pluginutils from dependencies

kazuma1989:
Hi. A small fix here 👍 

I've removed `@rollup/pluginutils` from the dependencies list because `createFilter` util is exposed from Vite v3. (vite >= 3 is in the peer deps, so it's safe)
Less dependencies are better.

- https://github.com/vitejs/vite/pull/8562
- #459

### DIFF
--- a/packages/builder-vite/package.json
+++ b/packages/builder-vite/package.json
@@ -14,7 +14,6 @@
   "homepage": "https://github.com/storybookjs/builder-vite/#readme",
   "dependencies": {
     "@joshwooding/vite-plugin-react-docgen-typescript": "0.0.5",
-    "@rollup/pluginutils": "^4.2.1",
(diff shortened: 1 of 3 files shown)


IanVS (approved):
Hm, this is one reason I'm not a huge fan of vite bundling all of their dependencies.  If they didn't do that, package managers could de-duplicate the `@rollup/pluginutils` that vite uses and the `@rollup/pluginutils` in the dependencies here, and there'd be no impact to the size of node_modules by us depending on it directly.  Instead, vite is stuck with having to re-export things and relying on other packages like this one to use what vite provides instead of using the dependencies directly.  

That said, it's a decision they've made, and now to avoid extra dependencies we do need this PR.  Thanks for submitting it.

PR #478, opened 2022-08-14
Replace InlineConfig with ResolvedConfig

bryanjtc:
I made a small improvement on the type `ViteFinal`. By changing `InlineConfig` with [`ResolvedConfig`](https://vitejs.dev/guide/api-javascript.html#resolvedconfig) most of the properties are resolved and non-undefined. This is better for type checking.


IanVS:
I'm not sure this is really the right approach though.  Many of the fields are, in fact, undefined in viteFinal.  We don't resolve the config by this point.  So, users might rely on a value being present, and it won't be.  

I think to be accurate, we would need to `Pick` the types that we are actually setting, which differ between production and development.

PR #482, opened 2022-08-18
Make sure mocked actions can be identified in production build

### LINKED ISSUE #206: Actions are not mocked for interaction tests in production build
After switching from webpack builder to this vite builder, the action args are not mocked correctly anymore.

Reproduction: https://github.com/bodograumann/storybook-vite-action-interaction

Use as:
```
git clone https://github.com/bodograumann/storybook-vite-action-interaction
cd storybook-vite-action-interaction
yarn install
yarn build-storybook
python -m http.server --directory storybook-static 6677
xdg-open http://localhost:6677/?path=/story/example-button--primary
```

Now the Interactions tab will be marked as failed with the message:
```
expect(received).toHaveBeenCalled()

Matcher error: received value must be a mock or spy function

Received has type:  function
Received has value: "[Function anonymous]"
```

In contrast, when running with `yarn storybook`, everything is fine.

bodograumann:
- Enable esbuild keep names as workaround
- Add interactive test in vue example

Not sure if the tests should be kept in, as they add a lot of dependencies to the example.

Fixes #206.


examples/vue/.storybook/main.ts
```diff
@@ -16,7 +16,10 @@ export default {
   async viteFinal(config, { configType }) {
     // Demonstrates use of mergeConfig and resolve.alias as an array
     return mergeConfig(config, {
-      resolve: { alias: [{ find: '@assets', replacement: resolve(__dirname, '..', 'stories', 'assets') }] },
+      resolve: {
+        alias: [{ find: '@assets', replacement: resolve(__dirname, '..', 'stories', 'assets') }],
+        dedupe: ['@storybook/client-api'],
```
IanVS, line 21: This `dedupe` is only necessary if there are different storybook versions being installed. I see that `@storybook/addon-interactions` doesn't match the rest, but from discord it sounded like this error started before you added that addon.  If you install `addon-interactions` 6.5.9, is this dedupe still needed?

bodograumann: I invistigated with `yarn why` and set `@storybook/vue3` as well as `@storybook/addon-interactions` to `6.5.9` as fixed version.This seems to work.

IanVS (approved):

IanVS:
Thanks @bodograumann!

PR #491, opened 2022-09-01
Rename Storybook root IDs for v7 compatibility

tjq:
Storybook v7 introduces a breaking change which renames the root elements that are selected during story rendering. This results in the following error when running `sb dev`:

```
TypeError: Cannot read properties of null (reading 'setAttribute')
    at WebView.showStory (http://localhost:6006/node_modules/.vite-storybook/deps/chunk-FSTL3CDP.js?v=bad8a373:1166:20)
    at WebView.prepareForStory (http://localhost:6006/node_modules/.vite-storybook/deps/chunk-FSTL3CDP.js?v=bad8a373:1075:10)
    at PreviewWeb.renderSelection (http://localhost:6006/node_modules/.vite-storybook/deps/chunk-FSTL3CDP.js?v=bad8a373:1609:52)
    at async PreviewWeb.selectSpecifiedStory (http://localhost:6006/node_modules/.vite-storybook/deps/chunk-FSTL3CDP.js?v=bad8a373:1408:5)
```

Which is triggered by this: https://github.com/storybookjs/storybook/blob/next/code/lib/preview-web/src/WebView.ts#L180-L183.

See https://github.com/storybookjs/storybook/pull/10638 for more information.

### DIFF
--- a/packages/builder-vite/input/iframe.html
+++ b/packages/builder-vite/input/iframe.html
@@ -22,8 +22,8 @@
   </head>
   <body>
     <!-- [BODY HTML SNIPPET HERE] -->
-    <div id="root"></div>
-    <div id="docs-root"></div>
+    <div id="storybook-root"></div>
+    <div id="storybook-docs"></div>
     <script type="module" src="/virtual:/@storybook/builder-vite/vite-app.js"></script>
   </body>
 </html>
--- a/packages/builder-vite/plugins/no-fouc.ts
+++ b/packages/builder-vite/plugins/no-fouc.ts
@@ -43,8 +43,8 @@ function insertHeadStyles(html: string) {
     :not(.sb-show-errordisplay) > .sb-errordisplay {
       display: none;
     }
-    #root[hidden],
-    #docs-root[hidden] {
+    #storybook-root[hidden],
+    #storybook-docs[hidden] {
       display: none !important;
     }
   </style>


IanVS:
Thanks for this!  The vite builder has actually been moved into the main storybook monorepo for v7, and at least part of the change was made already: https://github.com/storybookjs/storybook/blob/9d198e55fab5fa708b7d6045ca1f615f8f5c4982/code/lib/builder-vite/input/iframe.html#L25-L26.  But, the change to no-fouc was missed.  

But that reminds me that we can remove `no-fouc` entirely, since https://github.com/vitejs/vite/issues/6737 fixed it on vite's end, and we only support 3+ now.  I'll make that change over in the monorepo.  This package will be sticking around just for important bugfixes for storybook 6.5.  

Anyways, the good news is that if you're trying out storybook 7, you should be able to install @storybook/builder-vite at the same version as all the other main storybook packages.  The latest alpha also has a new feature which automatically uses your vite.config.js file as a base for the config.  It's early days, but would love to hear how it goes if you try it out!

IanVS:
Just to close the loop, see https://github.com/storybookjs/storybook/pull/19087 for some plugin cleanup, including removal of no-fouc.

tjq:
@IanVS thanks for the quick reply! I've installed `@storybook/builder-vite@7.0.0-alpha.28`, I originally did not see that it was now a part of the Storybook monorepo. I'm seeing some different errors now but I suspect that I'm better off starting over with the correct dependencies from the beginning 😄

IanVS:
If you hit any problems, please open an issue in the storybook repo, and tag me if you don't mind!

PR #505, opened 2022-09-29
feat: allow definition of stories in vue files

tobiasdiez:
I'm working on a [vite plugin](https://github.com/tobiasdiez/unplugin-storybook-vue) that transforms a vue sfc file to a standard Storybook CSF, so that one could specify stories in a format that is more convenient for vue users. This is inspired by [histoire](https://histoire.dev) and will implement https://github.com/storybookjs/storybook/issues/9768.

In this setting, stories are specified in `xyz.stories.vue` files, and then in a first step transformed to valid CSF js files. For this to work, the other vite plugins need to include or exclude these vue story files. This is done in this PR.

(As a follow-up, it would probably make sense to determine the files that the vite plugins operate on by looking at the `stories` config option in `.storybook/main.js`).

### DIFF
--- a/packages/builder-vite/code-generator-plugin.ts
+++ b/packages/builder-vite/code-generator-plugin.ts
@@ -39,7 +39,7 @@ export function codeGeneratorPlugin(options: ExtendedOptions): Plugin {
       // HMR to update the importFn.
       server.watcher.on('add', (path) => {
         // TODO maybe use the stories declaration in main
-        if (/\.stories\.(([tj]sx?)|(vue))$/.test(path) || /\.(story|stories).mdx$/.test(path)) {
+        if (/\.stories\.(?:[tj]sx?|vue)$/.test(path) || /\.(story|stories).mdx$/.test(path)) {
           // We need to emit a change event to trigger HMR
           server.watcher.emit('change', virtualStoriesFile);
         }
--- a/packages/builder-vite/code-generator-plugin.ts
+++ b/packages/builder-vite/code-generator-plugin.ts
@@ -39,7 +39,7 @@ export function codeGeneratorPlugin(options: ExtendedOptions): Plugin {
       // HMR to update the importFn.
       server.watcher.on('add', (path) => {
         // TODO maybe use the stories declaration in main
-        if (/\.stories\.([tj])sx?$/.test(path) || /\.(story|stories).mdx$/.test(path)) {
+        if (/\.stories\.(([tj]sx?)|(vue))$/.test(path) || /\.(story|stories).mdx$/.test(path)) {
           // We need to emit a change event to trigger HMR
           server.watcher.emit('change', virtualStoriesFile);
         }
--- a/packages/builder-vite/inject-export-order-plugin.ts
+++ b/packages/builder-vite/inject-export-order-plugin.ts
@@ -6,7 +6,7 @@ export const injectExportOrderPlugin = {
   // This should only run after the typescript has been transpiled
   enforce: 'post',
   async transform(code: string, id: string) {
-    if (!/\.stories\.([tj])sx?$/.test(id) && !/(stories|story).mdx$/.test(id)) {
+    if (!/\.stories\.(([jt]sx?)|(vue))$/.test(id) && !/(stories|story).mdx$/.test(id)) {
       return;
     }
     // TODO: Maybe convert `injectExportOrderPlugin` to function that returns object,
--- a/packages/builder-vite/inject-export-order-plugin.ts
+++ b/packages/builder-vite/inject-export-order-plugin.ts
@@ -6,7 +6,7 @@ export const injectExportOrderPlugin = {
   // This should only run after the typescript has been transpiled
   enforce: 'post',
   async transform(code: string, id: string) {
-    if (!/\.stories\.(([jt]sx?)|(vue))$/.test(id) && !/(stories|story).mdx$/.test(id)) {
+    if (!/\.stories\.(?:[tj]sx?|vue)$/.test(id) && !/(stories|story).mdx$/.test(id)) {
       return;
     }
     // TODO: Maybe convert `injectExportOrderPlugin` to function that returns object,
--- a/packages/builder-vite/plugins/vue-docgen.ts
+++ b/packages/builder-vite/plugins/vue-docgen.ts
@@ -7,7 +7,7 @@ export function vueDocgen(vueVersion: 2 | 3): Plugin {
     name: 'vue-docgen',
 
     async transform(src: string, id: string) {
-      if (/\.(vue)$/.test(id)) {
+      if (/\.(vue)$/.test(id) && !/\.(stories)\.(vue)$/.test(id)) {
         const metaData = await parse(id);
         const metaSource = JSON.stringify(metaData);
         const s = new MagicString(src);
--- a/packages/builder-vite/plugins/vue-docgen.ts
+++ b/packages/builder-vite/plugins/vue-docgen.ts
@@ -7,7 +7,7 @@ export function vueDocgen(vueVersion: 2 | 3): Plugin {
     name: 'vue-docgen',
 
     async transform(src: string, id: string) {
-      if (/\.(vue)$/.test(id) && !/\.(stories)\.(vue)$/.test(id)) {
+      if (/\.vue$/.test(id) && !/\.stories\.vue$/.test(id)) {
         const metaData = await parse(id);
         const metaSource = JSON.stringify(metaData);
         const s = new MagicString(src);
--- a/packages/builder-vite/source-loader-plugin.ts
+++ b/packages/builder-vite/source-loader-plugin.ts
@@ -3,7 +3,7 @@ import sourceLoaderTransform from '@storybook/source-loader';
 import type { ExtendedOptions } from './types';
 import MagicString from 'magic-string';
 
-const storyPattern = /\.stories\.[jt]sx?$/;
+const storyPattern = /\.stories\.(([jt]sx?)|(vue))$/;
 const storySourcePattern = /var __STORY__ = "(.*)"/;
 const storySourceReplacement = '--STORY_SOURCE_REPLACEMENT--';
 
--- a/packages/builder-vite/source-loader-plugin.ts
+++ b/packages/builder-vite/source-loader-plugin.ts
@@ -3,7 +3,7 @@ import sourceLoaderTransform from '@storybook/source-loader';
 import type { ExtendedOptions } from './types';
 import MagicString from 'magic-string';
 
-const storyPattern = /\.stories\.(([jt]sx?)|(vue))$/;
+const storyPattern = /\.stories\.(?:[jt]sx?|vue)$/;
 const storySourcePattern = /var __STORY__ = "(.*)"/;
 const storySourceReplacement = '--STORY_SOURCE_REPLACEMENT--';
 
--- a/packages/builder-vite/vite-config.ts
+++ b/packages/builder-vite/vite-config.ts
@@ -120,7 +120,7 @@ export async function pluginConfig(options: ExtendedOptions, _type: PluginConfig
   if (framework === 'vue3') {
     try {
       const vuePlugin = require('@vitejs/plugin-vue');
-      plugins.push(vuePlugin());
+      plugins.push(vuePlugin({exclude: [/\.stories\.vue$/]}));
     } catch (err) {
       if ((err as NodeJS.ErrnoException).code === 'MODULE_NOT_FOUND') {
         throw new Error(`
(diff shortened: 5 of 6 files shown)


tobiasdiez:
Thanks for your encouraging words and for the pointer to the svelte addon. I'm not very deep into the storybook environment, so its good to see how other projects are doing this. I learned a few good tricks while reading the code.

The storybook addon is a good idea and would simplify the installation for the user. Do you have an example of an addon modifying vite-builder's vite config?

>  In storybook 7.0, the vite builder will use the vite.config.js from the user, which means we won't be able to add an excludes like you've done here.

My issue was that without this exclude the transformation added by my vite plugin would have been overwritten by the vite/vue plugin. As there doesn't seem a possibility to stop a transformation chain in vite (from a plugin) and my plugin needs to be quite early in the chain, it was easiest to simple exclude the stories file from the vite/vue plugin. I guess it would also be possible (although not very nice) to instruct users to add the exclude mechanism to their vite config in v7 (or maybe do this automatically via the addon).

(Thanks for the invitation to discord, but I'm not a big fan of that platform. Would prefer to continue the discussion here or in https://github.com/tobiasdiez/unplugin-storybook-vue, or via email.)

IanVS:
>Do you have an example of an addon modifying vite-builder's vite config?

I don't think there are any addons which do this yet, but the frameworks do.  See https://github.com/storybookjs/storybook/blob/next/code/frameworks/vue3-vite/src/preset.ts#L22 for instance.

Note that this will run before the user's `viteFinal` (if they have one in their `.storybook/main.js`).  

>and my plugin needs to be quite early in the chain

Could you `unshift` the plugin onto the list of plugins, rather than `push`, so it ends up at the top?

tobiasdiez:
> > Do you have an example of an addon modifying vite-builder's vite config?
> 
> I don't think there are any addons which do this yet, but the frameworks do. See https://github.com/storybookjs/storybook/blob/next/code/frameworks/vue3-vite/src/preset.ts#L22 for instance.
> 
> Note that this will run before the user's `viteFinal` (if they have one in their `.storybook/main.js`).

Thanks, this looks simple enough. Is this a v7 specific feature or would it work with v.6.5 as well?
 
> > and my plugin needs to be quite early in the chain
> 
> Could you `unshift` the plugin onto the list of plugins, rather than `push`, so it ends up at the top?

That's what I'm currently doing, and its working in the sense that its executed before the other plugins from builder-vite. However, I do need to exclude stories files from the vite/vue plugin, which runs later and would otherwise overwrite the changes.

https://github.com/tobiasdiez/unplugin-storybook-vue/blob/3a8fed21f315bee55d324be2f9917c425ccef473/examples/vite/.storybook/main.ts#L25

tobiasdiez:
Short update from my side: The basic things are now working. In particular, I've started to add an one-to-one correspondence between the "classical" CSF format and the native vue format for the stories that are in the documentation, see https://github.com/tobiasdiez/unplugin-storybook-vue/tree/main/examples/vite/src/writing-stories.

Feedback on the current state is much appreciated. Where I'm currently stuck is the handling of args in a way that feels native to vue. I've added an initial proposal that is up for discussion at https://github.com/tobiasdiez/unplugin-storybook-vue/issues/13. What do you think?

IanVS:
@Dschungelabenteuer, @shilman I wonder if maybe you have some thoughts on this exciting work to add native vue stories.

tobiasdiez:
I've managed to find a workaround for the compatibility with the `vue` plugin, so the `exclude` hack is no longer necessary. The remaining changes of this PR are still needed to run or not run the plugin on the new `stories.vue` file.

It would be nice if this PR could be merged, so that I can release an early alpha for the vue-storybook addon that people can try out and give feedback on. 🚀

Dschungelabenteuer:
Sorry I didn't have time to take a deep dive into the plugin itself yet! However, there are a few thoughts and questions which cross my mind while taking a quick look at @tobiasdiez's work:

* From a Vite+Vue user's perspective, this looks promising and this could be a very interesting alternative way of writing stories.  Some of my colleagues are not quite familiar with Storybook and – despite the CSF format being pretty simple to use and get used to – it might prove more natural for them to write plain Vue SFCs.
* From the perspective of a potential user of this plugin, this raises two main questions for you @tobiasdiez. Sorry if you've already answered them before, I could not find this piece of information without going through your code:
  * Are you intending to make this work for both Vue 2 and 3? Vue 3 user here, but I know that for various reasons, its adoption is quite slower than expected among the Vue ecosystem, and that there is probably still a bunch of Storybook users relying on Vue 2. IIRC, while Storybook's Vite builder initially did not officially supported Vue 2, @IanVS worked on both Vue 2.6 and 2.7 support recently.
  * Since you went for an unified plugin, would the webpack5 builder be supported? My guess is, if it required some changes on Vite builder's side, it would also require some changes on the webpack one's?

* From a contributor's perspective, the actual changes of this PR made me wonder what would maintainers' take be here. As a maintainer, I would probably want to keep the codebase as much framework-agnostic as possible, and some of the changes here introduce a bit of framework specificity. Maybe some work could be done here to provide users with a way of customizing [patterns](https://github.com/storybookjs/storybook/blob/65957e6789bc57d506ca63e6f3be3927f110565e/code/lib/types/src/modules/core-common.ts#L379) and extensions used to match stories.

Anyway, this looks cool and I'll surely give it a try soon :)

tobiasdiez:
Thanks for the feedback! Concerning your points:
- Yes, the CSF format is easy enough to understand and write, but I'm really missing IDE support (Volar) when writing more complex stories/templates.
- I mostly focused on Vue 3 currently. But the code currently doesn't use anything specific to that version and the only dependency on vue is via the vue sfc compiler that is used to compile the template to a render function. So if the demand arises one could make the vue compiler a peer dependency and reuse whatever the user has installed.
- Webpack support is indeed planned, but I haven't yet looked at this. Since I'm not that familiar with the Storybook + Webpack combo, I would appreciate PRs in this direction. Shouldn't be too hard.
- I agree, it would be nice to refactor the whole filter mechanism to match the user-defined stories (and probably use vite's filter method https://github.com/rollup/plugins/tree/master/packages/pluginutils#createfilter).

tobiasdiez:
@IanVS Can you please review this PR? Thanks!

IanVS:
Hi, yes I will indeed find some time to review this, hopefully soon.

IanVS:
I haven't forgotten about this, just been busy lately!  Just wanted to let you know it's on my list of things to get to.

packages/builder-vite/plugins/vue-docgen.ts
```diff
@@ -7,7 +7,7 @@ export function vueDocgen(vueVersion: 2 | 3): Plugin {
     name: 'vue-docgen',
 
     async transform(src: string, id: string) {
-      if (/\.(vue)$/.test(id)) {
+      if (/\.(vue)$/.test(id) && !/\.(stories)\.(vue)$/.test(id)) {
```
IanVS, line 10: We should be able to ditch the capturing groups here.

```suggestion
      if (/\.vue$/.test(id) && !/\.stories\.vue$/.test(id)) {
```

packages/builder-vite/inject-export-order-plugin.ts
```diff
@@ -6,7 +6,7 @@ export const injectExportOrderPlugin = {
   // This should only run after the typescript has been transpiled
   enforce: 'post',
   async transform(code: string, id: string) {
-    if (!/\.stories\.([tj])sx?$/.test(id) && !/(stories|story).mdx$/.test(id)) {
+    if (!/\.stories\.(([jt]sx?)|(vue))$/.test(id) && !/(stories|story).mdx$/.test(id)) {
```
IanVS, line 9:
```suggestion
    if (!/\.stories\.(?:[tj]sx?|vue)$/.test(id) && !/(stories|story).mdx$/.test(id)) {
```

packages/builder-vite/code-generator-plugin.ts
```diff
@@ -39,7 +39,7 @@ export function codeGeneratorPlugin(options: ExtendedOptions): Plugin {
       // HMR to update the importFn.
       server.watcher.on('add', (path) => {
         // TODO maybe use the stories declaration in main
-        if (/\.stories\.([tj])sx?$/.test(path) || /\.(story|stories).mdx$/.test(path)) {
+        if (/\.stories\.(([tj]sx?)|(vue))$/.test(path) || /\.(story|stories).mdx$/.test(path)) {
```
IanVS, line 42:
```suggestion
        if (/\.stories\.(?:[tj]sx?|vue)$/.test(path) || /\.(story|stories).mdx$/.test(path)) {
```

packages/builder-vite/vite-config.ts
```diff
@@ -66,7 +66,7 @@ export async function pluginConfig(options: ExtendedOptions, _type: PluginConfig
     // We need the react plugin here to support MDX.
     viteReact({
       // Do not treat story files as HMR boundaries, storybook itself needs to handle them.
-      exclude: [/\.stories\.([tj])sx?$/, /node_modules/].concat(framework === 'react' ? [] : [/\.([tj])sx?$/]),
+      exclude: [/\.stories\.(([jt]sx?)|(vue))$/, /node_modules/].concat(framework === 'react' ? [] : [/\.([tj])sx?$/]),
```
IanVS, line 69:
```suggestion
      exclude: [/\.stories\.(?:[tj]sx?|vue)$/, /node_modules/].concat(framework === 'react' ? [] : [/\.([tj])sx?$/]),
```

packages/builder-vite/source-loader-plugin.ts
```diff
@@ -3,7 +3,7 @@ import sourceLoaderTransform from '@storybook/source-loader';
 import type { ExtendedOptions } from './types';
 import MagicString from 'magic-string';
 
-const storyPattern = /\.stories\.[jt]sx?$/;
+const storyPattern = /\.stories\.(([jt]sx?)|(vue))$/;
```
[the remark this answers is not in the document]
IanVS: If we change this storyPattern, we'll run the `sourceLoaderTransform` on .vue files, right?  IIUC, this is relying on your plugin running first, to convert the file from vue syntax to normal js (csf) syntax.  I wonder, could your addon / vite plugin change the `id` of the file, perhaps in the "resolveId" hook?  In that way, it might be possible to avoid bugs that could happen if someone writes a `.stories.vue` file without your addon.  Though, in that case it would just fail a different way, so maybe it doesn't matter.

packages/builder-vite/code-generator-plugin.ts
IanVS: @JReinhold, this makes me wonder about `.stories.svelte` files too.  This is a pattern we use to detect when we need to trigger HMR changes in stories files.  We have a similar set of regexes in 7.0 as well.

IanVS (commented):
Sorry this took me so long to get to.  I think this is fine to try out.  I made a small suggestion to improve the regexes, and I'm a little nervous about the source loader running on .vue files, but I'm willing to give this a shot and let you test things out further with your addon.  Mind updating the regexes, and I'll get this merged and released?  As long as @joshwooding doesn't have any concerns.

packages/builder-vite/code-generator-plugin.ts
JReinhold: huh, interesting.
I think the todo above:
> // TODO maybe use the stories declaration in main
is the right approach here, but in the mean time we could just have a regex that matches on `*.stories.*`? It would match on too much, but in this case I don't think that will be an issue?

packages/builder-vite/source-loader-plugin.ts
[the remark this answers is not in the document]
tobiasdiez: I tried to change the resolveId, also because it would make a few things easier in the plugin, but never got it to work. If I remember correctly, one of the issues was that the vue loader would always overwrite the resolve id again.

But as you say, it probably doesn't matter that much since it would only change the place where things are failing.

tobiasdiez:
Thanks for the review. I've now committed your suggested improvements to the regexes.

tobiasdiez:
Thanks you very much! I've now released a first early-preview of the addin at https://www.npmjs.com/package/storybook-vue-addon. Please feel free to play around with it if you have the time.

bodograumann:
This looks quite interesting. Have you thought about using the new native format as an alternative to writing stories and documentation in mdx format, @tobiasdiez ?

tobiasdiez:
These experiments are still at a very early phase and I've not thought about the integration with mdx/documentation. How is this handled by the native svelte addon? 
Maybe we could use the [MDC format](https://content.nuxtjs.org/guide/writing/mdc/) of nuxt/content to embed vue components and thus stories into markdown?

PR #509, opened 2022-10-05
Re-add typescript check when adding react-docgen-typescript plugin

joshwooding:
A hybrid of https://github.com/storybookjs/builder-vite/pull/460 and https://github.com/storybookjs/builder-vite/pull/313.

I believe this approach matches how storybook works with webpack.


IanVS (approved):
I think this makes sense.  Would you like to make the same change for the 7.0 vite-builder as well?

PR #531, opened 2022-12-16
Pin version of react-docgen

IanVS:
Fixes https://github.com/storybookjs/builder-vite/issues/530

The latest version of react-docgen has gone to ESM only, and made some other changes which break our usage.  This pins the version being used to a known-good version, which is safer during alphas anyway.


TomMcEachan:
This change is essential. Currently can't start a new storybook project because of this.

PR #548, opened 2023-01-21
Update @vitejs/plugin-react dependency for vite 4

IanVS:
Fixes https://github.com/storybookjs/builder-vite/issues/547

This also updates the version of mdx1-csf to one that no longer requires the use of the react vite plugin to compile jsx, it handles compilation itself.  So, we no longer need to include `@vitejs/plugin-react` in all projects, and react projects will need to make sure it's installed.  That makes this a breaking change.

And this also updates Vite to 4 in the examples, and fixes some missing dependencies.


IanVS:
@joshwooding, any idea what would be causing the parse errors in this PR?  I don't recall seeing this before.

IanVS:
Oh right, we had a problem with the original approach in storybook 7 as well when we tried to support Vite 4.  The results of sending MDX through the vite react-plugin still contain JSX at the end, whereas previously all the jsx was converted to javascript.  We never really figured out a solution to this in SB 7, we punted and said we'd only support MDX2, and we updated the way that mdx2-csf works so that it can handle jsx on its own (https://github.com/storybookjs/mdx2-csf/pull/30).  Maybe we need to do the same thing to mdx1-csf?  @ndelangen do you think that would be possible?  I also heard that SB 7 might need to continue supporting mdx-1 in an opt-in way, so I think that change to mdx2-csf might be needed regardless.

ndelangen:
Yes @IanVS @shilman and I looked at that this morning.. MDX1 support in 7.0 will make a come-back.

IanVS:
I'm not sure what's happening in the Workspaces example, something to do with the structure of the monorepo-within-a-monorepo, maybe.

With the latest changes, we can support mdx1 and mdx2.  I've also removed the addition of `@vitejs/plugin-react` from all projects since it is no longer required to process mdx.  This makes this a breaking change, since react users will need to be sure that the plugin is installed in their projects, just like for the other frameworks.

joshwooding (approved):

PR #552, opened 2023-02-03
Allow the use of @vitejs/plugin-react-swc

IanVS:
Fixes https://github.com/storybookjs/builder-vite/issues/551

This will first look for `@vitejs/plugin-react`, then `@vitejs/plugin-react-swc`.  It seems like the SWC version might not work with MDX stories, though.  I updated the react-18 example to use it, and the intro story doesn't load.  I don't think that should necessarily block this PR though.

### DIFF
--- a/examples/lit-ts/package.json
+++ b/examples/lit-ts/package.json
@@ -8,8 +8,8 @@
     "lit": "^2.2.2"
   },
   "scripts": {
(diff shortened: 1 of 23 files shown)


joshwooding (approved):

tastafur:
Thank you very much for your work @IanVS  : D

PR #556, opened 2023-02-07
Inject mdx compiler for mdx1

IanVS:
fixes https://github.com/storybookjs/builder-vite/issues/557

I found that mdx was not working correctly here, and @valentinpalkovic realized my fallback in Storybook 7 wasn't either (https://github.com/storybookjs/storybook/pull/20823).  It was because we used to inject the mdx compiler into the source code that we get back from the `@storybook/mdx1-csf` compiler, but we lost that in https://github.com/storybookjs/builder-vite/pull/548/files#diff-f2dfc4dfa9074d77a728a88e6629a1d66be8a0765cab8562526cd00fbae910e5L6.  This re-introduces it, with a bit better of a guarantee that the correct version is going to be loaded, by starting from inside `@storybook/mdx1-csf`.

I've also pinned mdx1-csf here, in case the import is moved from the loader to the compiler, as @shilman has suggested doing.  We'll need to adjust this if that happens.

### DIFF
--- a/examples/lit-ts/.storybook/main.cjs
+++ b/examples/lit-ts/.storybook/main.cjs
@@ -18,6 +18,16 @@ module.exports = {
     return mergeConfig(config, {
       // prettier-ignore
       plugins: [postcssLit({ include: ['**/*.scss', '**/*.scss\?*'] })],
+      // because rollup does not respect NODE_PATH, and we have a funky example setup that needs it
+      build: {
+        rollupOptions: {
+          plugins: {
+            resolveId: function (code) {
+              if (code === 'react') return require.resolve('react');
+            },
+          },
+        },
+      },
     });
   },
 };
--- a/examples/lit-ts/.storybook/main.cjs
+++ b/examples/lit-ts/.storybook/main.cjs
@@ -1,5 +1,6 @@
 const { mergeConfig } = require('vite');
 const postcssLit = require('rollup-plugin-postcss-lit');
+const slash = require('slash');
 
 module.exports = {
   stories: ['../stories/**/*.stories.mdx', '../stories/**/*.stories.@(js|jsx|ts|tsx)'],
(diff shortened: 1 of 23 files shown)


IanVS:
Hm, based on the CI failures we might need to also require react to be a peer dependency like we do in SB 7.  :-/

joshwooding:
Looks good. :)

IanVS:
Turns out that rollup does not respect NODE_PATH, so we need to do a bit of resolution hackery to get it to resolve react relative to the example, and not from within builder-vite's node_modules.  Normal projects will not need to worry about this, it's just due to our non-standard multi-project example structure.

IanVS:
@joshwooding I'm having trouble with Windows here.  Any chance you can give this a shot?

zhyd1997:
Hi @IanVS 

from [issue 21023](https://github.com/storybookjs/storybook/issues/21023)

CI passed when enabled `previewMdx2 `:

Screenshot 2023-02-10 at 18 52 37

https://github.com/zhyd1997/builder-vite/pull/2/files/5e198d79e680ddc8047238a8e8e6f0ffe83c2ab6..824c0585f12824f6f08a18f2d526dcfe81673987

so this is `mdx1-csf` transform issue:

https://github.com/storybookjs/builder-vite/blob/4d49af67e795330248c6967741eafb6c5f0cc6cf/packages/builder-vite/plugins/mdx-plugin.ts#L20-L26

IanVS:
Yes, mdx1 requires us to inject an import into the code, and the yarn portal we are using for the examples doesn't seem to be working very well, because imports in builder-vite are not resolving back to the example's node_modules without some help.  It's working fine on posix systems, but not windows. I'm tempted to merge anyway and hope that Josh can get our CI green later. It's only an issue in our examples I believe.

PR #573, opened 2023-04-13
Docs: Fix quotes in readme example

z9956:


IanVS:
Thanks for the catch!  Sorry, but it looks like prettier failed, it probably wants double quotes.

IanVS:
Looks like ci is failing for an unrelated reason.  Thanks for the contribution.

